PR #2, opened 2019-03-15
Feedback: Removed unused/unreachable code

danielkutik:
Removed two unused defines  and a unused value.
Further there is some unreachable code that needs to be looked at.

Do not merge, to be discussed.


andersson:
We should fix the missing unref of the udev objects, instead of just adding a FIXME - preferably as one patch removing the unused code and a separate that fixes the udev handling.

If you update your patch accordingly I would be very happy to merge it!

Thanks,
Bjorn

PR #11, opened 2020-05-04
feat: expose read and write timeouts to the user

ciband:
Added additional, optional CLI arguments to expose the timeouts
for reading and writing to the end user.  Hardcoded timeouts can
sometimes fail for large flashes.


andersson:
These timeouts should not relate to the user, but to the device and possibly what is being flashed - and we should generally not hit the timeouts. So let's instead correct the ones that are necessary in the implementation.

PR #21, opened 2021-05-06
firehose: fix >4GB lseek file offsets

flto:
The byte offset value for lseek can easily exceed 4GB, cast as off_t to
avoid overflow when converting sectors to bytes for lseek.

(and use lseek64/off64_t so this fixes 32-bit builds of qdl too)

### DIFF
--- a/firehose.c
+++ b/firehose.c
@@ -29,6 +29,7 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
+#define _FILE_OFFSET_BITS 64
 #include <sys/stat.h>
 #include <sys/time.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


firehose.c
```diff
@@ -29,6 +29,7 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  * POSSIBILITY OF SUCH DAMAGE.
  */
+#define _LARGEFILE64_SOURCE
```
andersson, line 32: According to feature_test_macros(7) this shouldn't be used in new code, but rather _FILE_OFFSET_BITS=64 is the way to go. Do you agree?

(Also, given that off_t is signed, don't we have the problem already at 31 bits - i.e. 2GB - files?)

flto:
Changed to use `_FILE_OFFSET_BITS`.

4GB is correct for 64-bit builds, the cast from unsigned to off_t lets it work up to 4GB

PR #25, opened 2021-08-03
show version number with argument --version and in usage printout

dmillard-TD:
I'd like to be able to do some rudimentary version control, and it's very handy to be able to query the version number.  This PR simply adds this capability.


andersson:
Resolved by PR #90

PR #29, opened 2022-03-22
qdl: Add support for UFS3.1 parameters

CosmicPenguin:
WriteBooster is a mandatory feature on UFS 3.1 and required for provisioning.
Add the XML fields for the firehose.

Signed-off-by: Jordan Crouse <jorcrous@amazon.com>


jwinarske:
I just ran a test on this.  In case where XML does not implement the new values, the parsing ends in error and never gets any further.  So this would be a regression for other scenarios.

quic-kdybcio:
Substituted by ad13228e6a591a82e7fb1a1f2459bbe15bc012b2

PR #32, opened 2022-07-08
qdl: Add a fix for read xml error

zixi11:
When xml has no last_sector, qdl will read xml failed. Add check
to see if last_sector exists


zixi11:
Hi Andersson:

Can this PR be merged?

Thanks.

PR #33, opened 2022-09-09
adding changes needed to support EMMC sector erasing

xerootg:
I implemented the erase command for EMMC (as well as an annoying hack for my target device). I also added couple tweaks for helping find errors when parsing the xml. I don't think this is very clean, atleast compared to the rest that you have, what would you like to see to make this something mergable?


quic-kdybcio:
Substituted by 33a2f151a63b839b388cf693263a39add70f45f5 (except for the odd quirk you described)

PR #37, opened 2023-02-17
Implement sparse image support (no deps on libsparse)

bmx666:
Hi @andersson, yep, it is another "sparse" version but without deps on libsparse. I've been testing it on several QCOM projects for over 2 years now, no problem.


parport0:
@bmx666 Hi,

Why is this PR a draft? Are there any flaws in this implementation?

bmx666:
Hi @parport0 

> Why is this PR a draft?

I'm waiting feedback from maintainer... 

> Are there any flaws in this implementation?

There are only two flaws:
* upload speed can drop significantly if the image contains a lot of small chunks, compared to QPST / QFIL. Some kind of optimization is needed to merge and send many small fragments at once, because image may contain several small fragments following each other.
* CHUNK_TYPE_FILL is not yet supported, but it can be easily implemented.

obbardc:
Looks pretty good to me, thanks @bmx666 
Slowdown is fine, I guess. But perhaps CHUNK_TYPE_FILL is more important ?

bmx666:
> Looks pretty good to me, thanks @bmx666 Slowdown is fine, I guess. But perhaps CHUNK_TYPE_FILL is more important ?

Hi @obbardc , I uploaded an update to support `CHUNK_TYPE_FILL` a long time ago. Also during my research on Qualcomm tools, they replace the 4-byte chunks of fill values with one big `CHUNK_TYPE_RAW`.
Anyway, I don't have access to any Qualcomm boards anymore, I couldn't confirm and verify this patch.

obbardc:
@bmx666 can you rebase this, please?

andersson:
Sorry for not commenting on this earlier. I think this PR looks good, but I wasn't able to validate the functionality - assuming PEBKAC I was going to get back to this, but didn't.

Given @bmx666 announcement that he doesn't have access to hardware it would be nice if someone with understanding of the sparse format/mechanism could conclude this effort (with appropriate credit to @bmx666 of course).

bmx666:
> @bmx666 can you rebase this, please?

Hi @obbardc , done! I didn't notice any serious conflicts, build the current branch on my side.

> if someone with understanding of the sparse format/mechanism could conclude this effort

@andersson  Sparse format is pretty simple
`sparse header + chunk header + (data optional) + chunk header + (data optional) + ...`

Sparse header contains version, checksum, size of non-sparse image and number of chunks.
Chunk header can be one of 3 types: RAW, FILL, DON'T CARE.

* **chunk header RAW** contains data size to RAW data + RAW data right after chunk header RAW.
* **chunk header FILL**  contains 4 bytes value + counter. Like flash 0x11223344 20 times.
* **chunk header DON'T CARE** contains how many bytes to skip (during QDL flashing process just increase EMMC/UFS cursor offset and process the next chunk)

If the XML does not contain the partition attribute `sparse="true"`, QDL will follow the default procedure. If the XML includes `sparse="true"` for a partition node but lacks a sparse header, it will be validated against the defined partition size. If the sizes match, it is likely that `sparse="true"` was set by mistake. Otherwise, if the size does not match the image size, QDL will return an error.
Qualcomm frequently ships an SDK where `sparse="true"` is mistakenly set, and some of their scripts used during image generation simply pass `sparse="true"` without actually generating a sparse image to final stage.

program.h
```diff
@@ -12,12 +12,16 @@ struct program {
 	const char *label;
 	unsigned num_sectors;
 	unsigned partition;
+	bool sparse;
```
obbardc, line 15:
```suggestion
	bool is_sparse;
```

nitpick: wonder if it makes sense to put these with `is_nand & is_erase` ?

sparse.c
```diff
@@ -0,0 +1,140 @@
+/*
+ * Copyright (c) 2016, Bjorn Andersson <bjorn@kryo.se>
```
obbardc, line 2: Is this copyright correct ? Surely it should be your name ?

sparse.h
```diff
@@ -0,0 +1,48 @@
+#ifndef __SPARSE_H__
```
obbardc, line 1: Does this header need a copyright ?

sparse.h
```diff
@@ -0,0 +1,48 @@
...
+	                         /* as 0. Standard 802.3 polynomial, use a Public Domain */
+	                         /* table implementation */
+} sparse_header_t;
+
+#define SPARSE_HEADER_MAGIC     0xed26ff3a
+#define SPARSE_HEADER_MAJOR_VER 0x0001
+#define SPARSE_HEADER_MINOR_VER 0x0000
+
+typedef struct __attribute__((__packed__)) chunk_header {
+	uint16_t chunk_type; /* 0xCAC1 -> raw; 0xCAC2 -> fill; 0xCAC3 -> don't care */
```
obbardc, line 25: thinking out loud; would an enum be better ?

obbardc (changes requested):
Need to add your Signed-off-by to commits too. There are just nitpicks in this review !

program.h
bmx666: Not sure, "sparse" naming reflects a real XML-attribute "sparse", same as "label", "num_sectors" and etc

sparse.c
bmx666: just copy-paste, I don't know what it's supposed to be

sparse.h
bmx666: I don't know what it's supposed to be.

sparse.h
bmx666: this is not a good idea because the compiler does not guarantee the size of enum, and using different optimization techniques enum can be 4 bytes or 1 byte, but never 2 bytes.

obbardc: No problem!

sparse.h
bmx666: The current file is based on original header from AOSP - https://android.googlesource.com/kernel/lk/+/5c761131621480d1288692a554805c8cc3fb5442%5E!/
Looks the same in the latest version - https://android.googlesource.com/platform/system/core/+/master/libsparse/sparse_format.h

sparse.c
bmx666: @ndechesne the current file was made by me, so I put my name in it

bmx666:
> Need to add your Signed-off-by to commits too. There are just nitpicks in this review !

@obbardc done!

program.c
```diff
@@ -80,6 +81,90 @@ static int load_erase_tag(xmlNode *node, bool is_nand)
...
+	struct program *program_sparse = NULL;
+	struct program *programes_sparse = NULL;
+	struct program *programes_sparse_last = NULL;
+	char tmp[PATH_MAX];
+
+	sparse_header_t sparse_header;
+	unsigned start_sector, chunk_size, chunk_type, chunk_data;
+
+	if (sparse_header_parse(fd, &sparse_header)) {
+		/* ignore false sparse images */
```
andersson, line 95: Isn't this comment misleading? If I read the code correctly we'll be programming the referenced image as a normal (non-sparse) image, not ignoring it.

program.c
```diff
@@ -80,6 +81,90 @@ static int load_erase_tag(xmlNode *node, bool is_nand)
...
+			return program_sparse;
+		}
+
+		fprintf(stderr, "[PROGRAM] Unable to parse sparse header at %s...failed\n",
+				program->filename);
+		return NULL;
+	}
+
+	for (uint32_t i = 0; i < sparse_header.total_chunks; ++i) {
+
```
andersson, line 110: As you're in this function fixing anyways, could you please drop this blank line?

firehose.c
```diff
@@ -411,19 +415,38 @@ static int firehose_program(struct qdl_device *qdl, struct program *program, int
...
+			case CHUNK_TYPE_RAW:
+				lseek(fd, (off_t) program->sparse_chunk_data, SEEK_SET);
+				break;
+			case CHUNK_TYPE_FILL:
+				fill_value = (uint32_t) program->sparse_chunk_data;
+				for (n = 0; n < max_payload_size; n += sizeof(fill_value))
+					memcpy(buf + n, &fill_value, sizeof(fill_value));
+				break;
+			default:
+				fprintf(stderr, "[SPARSE] invalid chunk type\n");
```
andersson, line 431: ux_err("invalid sparse chunk type\n");

program.c
```diff
@@ -80,6 +81,90 @@ static int load_erase_tag(xmlNode *node, bool is_nand)
...
+		/* ignore false sparse images */
+		if (program->sector_size * program->num_sectors == lseek(fd, 0, SEEK_END)) {
+			program_sparse = calloc(1, sizeof(struct program));
+			memcpy(program_sparse, program, sizeof(struct program));
+			program_sparse->sparse = false;
+			program_sparse->next = NULL;
+			return program_sparse;
+		}
+
+		fprintf(stderr, "[PROGRAM] Unable to parse sparse header at %s...failed\n",
```
andersson, line 104: ux_err() please.

program.c
```diff
@@ -80,6 +81,90 @@ static int load_erase_tag(xmlNode *node, bool is_nand)
...
+
+		chunk_type = sparse_chunk_header_parse(fd, &sparse_header, &chunk_size, &chunk_data);
+
+		switch (chunk_type) {
+			case CHUNK_TYPE_RAW:
+			case CHUNK_TYPE_FILL:
+			case CHUNK_TYPE_DONT_CARE:
+				break;
+			default:
+				fprintf(stderr, "[PROGRAM] Unable to parse sparse chunk %i at %s...failed\n",
```
andersson, line 119: ux_err() please.

program.c
```diff
@@ -80,6 +81,90 @@ static int load_erase_tag(xmlNode *node, bool is_nand)
...
+						i, program->filename);
+				return NULL;
+		}
+
+		if (chunk_size == 0)
+			continue;
+
+		if (chunk_size % program->sector_size != 0) {
+			fprintf(stderr, "[SPARSE] File chunk #%u size %u is not a sector-multiple\n",
```
andersson, line 128: ux_err() please.

program.c
```diff
@@ -192,7 +278,21 @@ int program_execute(struct qdl_device *qdl, int (*apply)(struct qdl_device *qdl,
 			continue;
 		}
 
-		ret = apply(qdl, program, fd);
+		if (!program->sparse) {
+			ret = apply(qdl, program, fd);
+		} else {
+			program_sparse = program_load_sparse(program, fd);
+			if (!program_sparse) {
+				fprintf(stderr, "[PROGRAM] load sparse failed\n");
```
andersson, line 286: I believe you printed an (more useful) error in each error path within program_load_sparse() already, so no need to tell the user once more.

sparse.c
```diff
@@ -0,0 +1,140 @@
...
+#include <unistd.h>
+
+#include "sparse.h"
+
+int sparse_header_parse(int fd, sparse_header_t* sparse_header)
+{
+	lseek(fd, 0, SEEK_SET);
+
+	if (read(fd, sparse_header, sizeof(sparse_header_t)) != sizeof(sparse_header_t)) {
+		fprintf(stderr, "[SPARSE] Unable to read sparse header\n");
```
andersson, line 49: Please convert each one of these to ux_err()

program.c
bmx666: updated to

```
/*
 * If the XML tag "program" contains the attribute 'sparse="true"'
 * for a partition node but lacks a sparse header,
 * it will be validated against the defined partition size.
 * If the sizes match, it is likely that the 'sparse="true"' attribute
 * was set by mistake.
 */
```

program.c
bmx666: removed

firehose.c
bmx666: done

program.c
bmx666: done

program.c
bmx666: done

program.c
bmx666: done

sparse.c
bmx666: done

program.c
bmx666: Hmm, you can generate a correct sparse image, but all chunks will be DON'T CARE or the chunk header will contain zero bytes of data.
On the one hand, the sparse image is validated, but on the other hand, there is nothing to flash and the program will have to be skipped.
Maybe you can suggest a better scenario?

igoropaniuk:
@bmx666 Could you please address all CI-reported issues as well, including checkpatch warnings and errors?

bmx666:
@igoropaniuk omg, did someone verify checker on false-positive errors and user-friendly for human reading before integration into mainline? Like remove space between type conversation that leads to completely unreadable code. It's seriously hard to read now - I have dyslexia, and removing spaces just makes it worse to read. Nowadays, disk space isn't that expensive.

andersson:
As the project has moved to SPDX header format and the indentation warnings looks valid, so please fix them.

All the errors about uint32_t vs u32, camelcase and typedefs are useless, please ignore them - they should not be part of the integration criteria (i.e. the workflows)!

The Windows build error needs to be addressed as well.

bmx666:
> As the project has moved to SPDX header format and the indentation warnings looks valid, so please fix them.

Is it okay to add SPDX header to sparse.h that belongs to Android under Apache 2.0 license?

> All the errors about uint32_t vs u32, camelcase and typedefs are useless, please ignore them - they should not be part of the integration criteria (i.e. the workflows)!

Ok, got it.

> The Windows build error needs to be addressed as well.

Sparse implementation depends on byte order functions like `ntohs` and `ntohl`. I fixed header to include `winsock2.h` for Windows platform but seems Windows needs to link library "ws2_32" and it must be added into Makefile.

bmx666:
I added workaround for Windows platform in Makefile to check if `OS` env is equal to `Windows_NT` and then link with `ws2_32`, but I'm not sure what is the best option.

igoropaniuk:
@bmx666 thanks for reporting all these false positives, I've addressed that in https://github.com/linux-msm/qdl/pull/116 (comments are welcome) and extended the list of rules to be ignored

sparse.c
```diff
@@ -0,0 +1,144 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright (c) 2025, Maksim Paimushkin <maxim.paymushkin.development@gmail.com>
+ * All rights reserved.
+ *
```
andersson, line 5: Keep the Copyright line (and All rights reserved), the remainder of this notice is "abbreviated" through the SPDX line above and can be dropped.

sparse.h
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: BSD-3-Clause
```
andersson, line 1: Per https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/license-rules.rst#n71 the SPDX line in header files should be /* */ comments instead of //.

More importantly thought, this file is Apache 2.0, so I believe this line should be:
/* SPDX-License-Identifier: Apache-2.0 */

sparse.h
```diff
@@ -0,0 +1,76 @@
...
+/*
+ * Copyright (C) 2010 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
andersson, line 5: As above, the Copyright line should be retained but this part should be dropped.

andersson (changes requested):
I'm sorry, but one more round of fixes to the license headers and then I think we're good.

JohnSagaQuic:
Revisited the topic.  We generated a test build with this change and verify on a QRD8650 device. After flashing, the device cannot bootup. It looks like need to add code to avoid repeated reading of fill chunk value.

PR #39, opened 2023-03-05
qdl: Allow user to decide USB OUT buffer multiplier

WillDoItMyself:
Due to issue related in 760b3df qdl is not sending big chunk of data at once to USB driver. That affects performance of flashing. 

This change allows user to set multiplier of buffer size. If may cause issues on some machines but gives huge speed gain.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 OUT := qdl
 
-CFLAGS := -O2 -Wall -g `xml2-config --cflags`
-LDFLAGS := `xml2-config --libs` -ludev
+CFLAGS := -O2 -Wall -g `pkg-config --cflags libxml-2.0`
+LDFLAGS := `pkg-config --libs libxml-2.0 libudev`
 prefix := /usr/local
 
 SRCS := firehose.c qdl.c sahara.c util.c patch.c program.c ufs.c
--- a/README
+++ b/README
@@ -14,3 +14,29 @@ e.g. the libxml2-dev package.
 
 With this installed run:
   make
+
+Buffer multiplier
+========
+**Warning**
+Using this parameter might not work stable on your system. Read 760b3df for
+more details. 
+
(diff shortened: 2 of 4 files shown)


bmx666:
Very useful feature and improvement, it saves me a lot of time, instead of 15 minutes I can flash devices in 2 minutes max! Thank you!

z3ntu:
Can confirm this works quite well. With this flashing goes from ~14MiB/s without this PR to about ~44MiB/s with `--multiplier 2048`

|multiplier|flashing speed|
|---|---|
|128|38262kB/s|
|256|40948kB/s|
|512|42437kB/s|
|1024|43222kB/s|
|2048|44038kB/s|

When rebasing on master there's some minor conflicts, `struct qdl_device` has moved to the file `qdl.h`

andersson:
Looking at that problem description again, I am puzzled to the exact details and why this wasn't fixed in the kernel instead - it seems unreasonable to allow user space to DOS the kernel like that. And indeed, there is a comment in the kernel documenting that the length can be "almost arbitrarily large". I also don't understand why I accepted changing this to 512 bytes.

With that in mind, I think the proposed multiplier-based mechanism favors user control over user friendliness, and I instead suggest that we clamp the value to e.g 128KB and if that breaks people's systems we 1) debug the kernel 2) consider our options for knobs etc.

andersson:
Replaced with #73, thanks for pushing for this!

z3ntu:
It's taken me a while to test the new version, but flashing without any extra arguments now take 2.6 minutes for my device, with flashing speed for super partition being around 41000kB/s to 42000kB/s, so being very close to what I tested at https://github.com/linux-msm/qdl/pull/39#issuecomment-2061451681, and I have a bit different USB connections now compared to when I last tested.

Thanks @andersson!

PR #43, opened 2023-05-09
Linter Fixes

jwinarske:
Addresses all clang-tidy complaints


andersson:
There seem to be some improvements in this patch, but the change of indentation everywhere hides this in noise and make the patch un-mergable.

PR #44, opened 2023-05-09
CMake

jwinarske:


konradybcio:
We're moving all linux-msm projects to meson, sorry.

quic-kdybcio:
BTW you may also be interested in https://github.com/qualcomm/qdlrs/, which I hope we'll gradually try to migrate towards..

PR #53, opened 2024-01-26
Add support for <read> operations

cpfair:
These operations follow a similar syntax to `<program>` operations - just moving data in the other direction.

To avoid shadowing the `read` function, I used `read_op` for types/functions/variables (breaking the existing convention).

Tested on a QCS6490 SoC.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ CFLAGS := -O2 -Wall -g `pkg-config --cflags libxml-2.0`
 LDFLAGS := `pkg-config --libs libxml-2.0 libudev`
 prefix := /usr/local
 
-SRCS := firehose.c qdl.c sahara.c util.c patch.c program.c ufs.c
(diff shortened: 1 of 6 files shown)


cpfair:
`<read ...>` operations in QFIL XML files are found in the wild (e.g. [this github search](https://github.com/search?q=%22%3Cread%22+%22start_sector%22+language%3Axml&type=code)). This PR was intended to make files like these work with `qdl`.

I agree that a `qdl-read` tool would be a useful addition. Perhaps it could be as simple as a `qdl --read` flag, which takes a regular QFIL XML file but performs reads instead of writes.

andersson:
I didn't want to include the merge commit in the history, and the code needed a few small fixups (missing stdlib.h and CodeQL complained about the transfer-speed-print). So rebased it, fixed those and merged it in the form of #74. Thank you for your contribution!

PR #56, opened 2024-03-26
Introduce ramdump support

quic-bjorande:


.gitignore
```diff
@@ -1,3 +1,5 @@
 *.o
 qdl
 ks
+compile_commands.json
```
lumag, line 4: What is LSP?

sahara.c
```diff
@@ -284,14 +300,124 @@ static int sahara_done(struct qdl_device *qdl, struct sahara_pkt *pkt)
...
+		read_req.length = SAHARA_MEM_READ64_LENGTH;
+		read_req.debug64_req.addr = region.addr + chunk;
+		read_req.debug64_req.length = remain;
+		n = qdl_write(qdl, &read_req, read_req.length);
+		if (n < 0)
+			break;
+
+		offset = 0;
+		while (offset < remain) {
+			n = qdl_read(qdl, buf, DEBUG_BLOCK_SIZE, 30000);
```
lumag, line 340: what is 30000?

sahara.c
```diff
@@ -284,14 +300,124 @@ static int sahara_done(struct qdl_device *qdl, struct sahara_pkt *pkt)
...
+	size_t chunk;
+	ssize_t n;
+	void *buf;
+	int fd;
+
+	buf = malloc(DEBUG_BLOCK_SIZE);
+	if (!buf)
+		return -1;
+
+	fd = openat(ramdump_dir, region.filename, O_WRONLY | O_CREAT, 0644);
```
lumag, line 320: I wonder if it is easier to mmap the fd and then just read to the mmap'ed area? Then we don't have to do the dance around the malloc/free/pointers/etc.

qdl.c
```diff
@@ -391,6 +391,9 @@ static void print_usage(void)
 	fprintf(stderr,
 		"%s [--debug] [--storage <emmc|nand|ufs>] [--finalize-provisioning] [--include <PATH>] <prog.mbn> [<program> <patch> ...]\n",
 		__progname);
+	fprintf(stderr,
+		"%s --ramdump [segment-filter,...] [--ramdump-path <PATH>]\n",
+		__progname);
```
lumag, line 396: I think, this looks like a completely separate tool from the UX point of view. You can not mix download and rumdump at the same time, can you?

lumag (changes requested):

sahara.c
andersson: That is a long timeout... I had some timeout issues during testing and bumped it, seems reasonable that we bring it down slightly...

qdl.c
andersson: Correct, you would either download or ramdump at any point in time, as such I did consider implementing it as a separate binary. But it's two modes of a very similar thing, so it felt reasonable to associate them both with the same command, and I didn't want to clutter people's /usr/local/bin unnecessarily.

lumag: I'd still suggest two commands. It sounds very strange to call the same binary in this way. But maybe it's just me :-D

sahara.c
lumag: I see. ack then.

.gitignore
lumag: I see. A purist in me says that this is irrelevant to the ramdump, but who cares :-D

andersson: Ahh, you're right. I forgot that we (typically) end up with a merge commit when accepting the PR. Too long since I contributed outside the kernel...

qdl.c
andersson: I associate the two states/operations and hence find it convenient to only have to have one tool laying around for mucking with that (and only one tool to remember the name of). I'd expect/hope distros pack both binaries, so that the user have them both...

What do you want the ramdump binary to be called then?

sahara.c
```diff
@@ -246,14 +301,149 @@ static int sahara_done(struct qdl_device *qdl, struct sahara_pkt *pkt)
...
+
+		offset = 0;
+		while (offset < remain) {
+			n = qdl_read(qdl, buf, DEBUG_BLOCK_SIZE, 30000);
+			if (n < 0) {
+				warn("failed to read ramdump chunk");
+				goto out;
+			}
+
+			write(fd, buf, n);
```
ahalaney, line 346: I built this real quick today (in hopes of testing it, but that's looking like another day now lol), and noticed this:

```
# make
cc -O2 -Wall -g `pkg-config --cflags libxml-2.0`   -c -o qdl.o qdl.c
cc -O2 -Wall -g `pkg-config --cflags libxml-2.0`   -c -o sahara.o sahara.c
sahara.c: In function ‘sahara_debug64_one’:
sahara.c:346:4: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
    write(fd, buf, n);
    ^~~~~~~~~~~~~~~~~
```

qdl.c
konradybcio: qramdump

PR #57, opened 2024-04-17
program: fail correctly on unrecognized tags

z3ntu:
Previously the message would be printed that the tag gets ignored but in reality the 'errors' variable will still be initialized as -EINVAL so the if below would return an error.

    [PROGRAM] unrecognized tag "zeroout", ignoring
    qdl: program_load rawprogram0.xml failed

Fix this.

### DIFF
--- a/program.c
+++ b/program.c
@@ -127,7 +127,7 @@ int program_load(const char *program_file, bool is_nand)
 	xmlNode *node;
 	xmlNode *root;
 	xmlDoc *doc;
-	int errors;
+	int errors = 0;
 
 	doc = xmlReadFile(program_file, NULL, 0);
 	if (!doc) {
@@ -140,8 +140,6 @@ int program_load(const char *program_file, bool is_nand)
 		if (node->type != XML_ELEMENT_NODE)
 			continue;
 
-		errors = -EINVAL;
-
 		if (!xmlStrcmp(node->name, (xmlChar *)"erase"))
 			errors = load_erase_tag(node, is_nand);
 		else if (!xmlStrcmp(node->name, (xmlChar *)"program"))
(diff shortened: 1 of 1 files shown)


lumag:
Should qdl instead fail on unrecognized tags?

konradybcio:
hm, on a second thought some unsupported/unrecognized tags may be "important".. perhaps it could use a --flag?

z3ntu:
This is for "zeroout" and "information" tags. The instructions we got from the ODM was to just remove them from the file for Qualcomm's fh_loader, so I also wouldn't be sad to just fail on unrecognized tags.

But the message is there right now and the code doesn't do what the print is saying, that's why I opened this ;)

z3ntu:
I've edited my previous comment for clarity, I think I wrote it a bit unclear.

To be clear, I have no idea what tool would handle "zeroout" (I'm guessing this tag is meant to just write zeroes to the partition?). `<information` sounds safer to ignore, I'm okay with ignoring that one.

But I can look into making this fail on purpose on unsupported tags, should be only a few lines change at max anyways.

lumag (changes requested):
I think it would be better to handle known known tags (even if by ignoring them) and break on on the unknown tags.

z3ntu:
Not sure what I can "handle" there, also not sure we should just ignore them.

These are the tags from rawprogram0.xml that our partner suggests to just remove from the file for QFIL flashing.
```xml
<zeroout SECTOR_SIZE_IN_BYTES="4096" factory_erase_only="false" label="userdata" num_partition_sectors="0" physical_partition_number="0" start_sector="3790029"/>
<zeroout SECTOR_SIZE_IN_BYTES="4096" factory_erase_only="false" label="metadata" num_partition_sectors="7648" physical_partition_number="0" start_sector="8736"/>
<information FCM_VERSION="1" STORAGE_TYPE="ufs" TABLE_VERSION="2"/>
```
(Similar `<information>` tags in other rawprogram*.xml files are present also)

I think I'll just rework this PR to just properly disallow unrecognized tags and leave the rest for someone else to handle.

lumag:
Sounds good.

z3ntu:
Implemented the change. Also added a commit on top to clean up some error handling there. Let me know what you think.

PR #58, opened 2024-05-03
Makefile: Generate compile_commands.json

quic-bjorande:
Introduce a make target for generating compile_commands.json to feed clangd and the Language Server Protocol, to gain better editor integration.

Add the relevant files to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 *.o
 qdl
 ks
+compile_commands.json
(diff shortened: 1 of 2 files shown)


konradybcio:
sneaked in #56

andersson:
That is scary. I dropped the patch from #56 and posted this separately, did not expect Github to merge it there...

PR #59, opened 2024-05-08
Transition to libusb and support targeting device by serial number

quic-bjorande:
Transitioning to libusb avoids open-coded USB descriptor parsing and removes the strong dependency
on the Linux-specific ioctl-based interface.
This should allow QDL to run on other OSs, it cleans up the code a bit, and it allow us to more conveniently
implement reading of the string descriptors, for a serial number comparison, in the support for selecting a
specific board when more than one is available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 QDL := qdl
 RAMDUMP := qdl-ramdump
 
-CFLAGS := -O2 -Wall -g `pkg-config --cflags libxml-2.0`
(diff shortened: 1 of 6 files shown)


ricardosalveti:
That's awesome, will give it a try as well, thanks for doing the work!

ricardosalveti:
Tested flashing lmp/yocto image on a mac m3 and worked fine!

PR #64, opened 2024-05-30
added kernel detach to avoid error: LIBUSB_ERROR_BUSY

AMV007:


AMV007:
*fixed kernel claim
*increased timeout for bulk transfer, not enough 100 ms

firehose.c
```diff
@@ -130,13 +130,13 @@ static int firehose_read(struct qdl_device *qdl, int timeout_ms,
 	timeradd(&now, &delta, &timeout);
 
 	for (;;) {
-		n = qdl_read(qdl, buf, sizeof(buf), 100);
+		n = qdl_read(qdl, buf, sizeof(buf), 700);
```
andersson, line 133: But when we hit this timeout, won't timercmp() below result in the loop to be continued?

usb.c
```diff
@@ -193,8 +193,10 @@ int qdl_read(struct qdl_device *qdl, void *buf, size_t len, unsigned int timeout
 	int ret;
 
 	ret = libusb_bulk_transfer(qdl->usb_handle, qdl->in_ep, buf, len, &actual, timeout);
-	if (ret < 0)
+	if (ret < 0){
+		warnx("bulk read failed: %s", libusb_strerror(ret));
```
andersson: Will this not result in two timeout prints when a timeout happens?
And for the initial firehose_read() in firehose_run(), we do expect to get a timeout once we've consumed the LOG entries. I would hope we can remove the existing "firehose operation timed out" print.

andersson:
We did some testing and have confirmed that the LIBUSB_ERROR_BUSY is something we want, so I was hoping to merge that.
Could you please provide a small problem description and split out the timeout change to its own PR?

firehose.c
quic-bjorande: Please see pull request #67.

AMV007:
ok, created separate commit only for LIBUSB_ERROR_BUSY : https://github.com/linux-msm/qdl/pull/68

I got mate, who can reflash over Windows with QFIL without problems, but at linux got constant timeout in random places, usually at big image part 2 gb:

<data><program SECTOR_SIZE_IN_BYTES="4096" num_partition_sectors="542383" physical_partition_number="0" start_sector="9480" filename="super_3.img"/></data>

FIREHOSE READ: <?xml version="1.0" encoding="UTF-8" ?>
<data>
<log value="INFO: Calling handler for program" /></data>
LOG: INFO: Calling handler for program
FIREHOSE READ: <?xml version="1.0" encoding="UTF-8" ?>
<data>
<response value="ACK" rawmode="true" /></data>
ERROR: n = -1, errno = 110 (Connection timed out)
qdl: failed to write: Connection timed out

PR #67, opened 2024-06-07
usb: Correct libusb timeout handling

quic-bjorande:
The documentation for libusb_bulk_transfer() specifies that in the event that the read or write request is split into multiple chunks, some of these may still be transferred despite a LIBUSB_ERROR_TIMEOUT is returned.

In the transition to libusb this detail was missed and completed read transfers are sometimes considred to be timeouts and the data discarded, obviously resulting in failure to continue.

Consider the "transferred" value in the event of timeout, to avoid this.

Although not yet spotted in testing, the same logic is introduced for the write path.


AMV007:
ok, created separate commit only for LIBUSB_ERROR_BUSY : https://github.com/linux-msm/qdl/pull/68

PR #69, opened 2024-06-07
*added version info and build time info

AMV007:
*added vscode settings to exclude
*max_payload_size to more clear size

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,10 @@
 QDL := qdl
 RAMDUMP := qdl-ramdump
 
-CFLAGS += -O2 -Wall -g `pkg-config --cflags libxml-2.0 libusb-1.0`
+BUILD_TIME := $(shell date +"%c")
+VERSION := "1.0.0"
+
+CFLAGS += -O2 -Wall -g `pkg-config --cflags libxml-2.0 libusb-1.0` -D "BUILD_TIME=\"$(BUILD_TIME)\"" -D "VERSION=\"$(VERSION)\""
 LDFLAGS += `pkg-config --libs libxml-2.0 libusb-1.0`
 prefix := /usr/local
 
--- a/firehose.c
+++ b/firehose.c
@@ -192,7 +192,7 @@ static int firehose_write(struct qdl_device *qdl, xmlDoc *doc)
 	return ret < 0 ? -saved_errno : 0;
 }
 
-static size_t max_payload_size = 1048576;
+static size_t max_payload_size = 1024*1024;
 
 /**
  * firehose_configure_response_parser() - parse a configure response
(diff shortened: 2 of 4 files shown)


konradybcio:
couple points:

- this is all good improvements
- most of the commiters here are used to the linux kernel way of structuring changes (one change per commit), but that's not a big deal for me with such small patches
- qdl (lowercase) seems to be the used throughout the codebase
- I think having a git hash in the print would be very useful (say `qdl version 1.0.0-1234abde`)
- the `date` could use a `LC_ALL=C` in the front
- since i added so many nits, I may as well ask to put a space between the `*` and the operands in the `max_payload_size` declaration :wink: 

thanks!

Makefile
```diff
@@ -1,7 +1,10 @@
 QDL := qdl
 RAMDUMP := qdl-ramdump
 
-CFLAGS += -O2 -Wall -g `pkg-config --cflags libxml-2.0 libusb-1.0`
+BUILD_TIME := $(shell date +"%c")
```
Rondom, line 4: This makes the build unreproducible. Should use SOURCE_DATE_EPOCH, see https://reproducible-builds.org/docs/source-date-epoch/

Makefile
```diff
@@ -1,7 +1,10 @@
...
+VERSION := "1.0.0"
```
Rondom, line 5: Probably good to add git-version info (git describe) as well or instead.

Makefile
quic-bjorande: Why is this information valuable. Please use the commit message to describe your motivation.

Makefile
quic-bjorande: For the version information to be useful (e.g. for bug reports etc), we need to bump this version string on every commit. So, I think git describe provides much more value here.

firehose.c
```diff
@@ -192,7 +192,7 @@ static int firehose_write(struct qdl_device *qdl, xmlDoc *doc)
 	return ret < 0 ? -saved_errno : 0;
 }
 
-static size_t max_payload_size = 1048576;
+static size_t max_payload_size = 1024*1024;
```
quic-bjorande, line 195: Unrelated change

qdl.c
```diff
@@ -111,6 +111,7 @@ int main(int argc, char **argv)
 	int opt;
 	bool qdl_finalize_provisioning = false;
 
+	printf("QDL version %s %s\n", VERSION, BUILD_TIME);
```
quic-bjorande, line 114: I'd like to make the tool less noisy... How about instead adding a --version argument which prints the version and exits?

quic-bjorande (changes requested):

Makefile
AMV007: sure, as soon as I will debug non working qdl at the PC of my colleague may be I will return to this change ;-)

andersson:
Resolved by PR #90

PR #70, opened 2024-06-07
added kernel detach to avoid error: LIBUSB_ERROR_BUSY

AMV007:
fix kernel detach error, for some reason can't use qdl without this at my pc

### DIFF
--- a/usb.c
+++ b/usb.c
@@ -126,9 +126,16 @@ static int qdl_try_open(libusb_device *dev, struct qdl_device *qdl, const char *
 			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


quic-bjorande:
I did test this and similar to the timeout, libusb return codes are tricky. So, checking for negative return value will break the case where the kernel doesn't have a driver attached to the USB device. I took the liberty of posting #71 which handles this, and with a commit message documenting why...

konradybcio:
Superseded by #71, hope it's okay

AMV007:
nice, thank you ;)

PR #72, opened 2024-06-07
For linux msm/firehose silence timeout warning

quic-bjorande:


firehose.c
```diff
@@ -53,6 +53,11 @@
 #include "qdl.h"
 #include "ufs.h"
 
+enum {
+	FIREHOSE_ACK,
```
konradybcio, line 57: "Clean this up by defining ACK and NAK in the positive value space and
use the negative value space for errno."

almost, but I don't mind ;)

maybe add a =value to the first entry

firehose.c
```diff
@@ -106,10 +111,15 @@ static int firehose_generic_parser(xmlNode *node, void *data)
...
-		return 0;
+		return -EAGAIN;
 	}
 
-	return xmlStrcmp(value, (xmlChar*)"ACK") == 0 ? 1 : -1;
+	if (xmlStrcmp(value, (xmlChar*)"ACK") == 0)
+		return FIREHOSE_ACK;
+	if (xmlStrcmp(value, (xmlChar*)"NAK") == 0)
+		return FIREHOSE_NAK;
+
```
konradybcio, line 121: should we print whatever came out here?

andersson: I think we should follow up and adjust the error prints throughout the file to give better error messages. Perhaps strerror(ret) vs "device rejected" printing.

firehose.c
```diff
@@ -267,14 +277,18 @@ static int firehose_configure(struct qdl_device *qdl, bool skip_storage_init, co
...
+	}
 
 	/* Retry if remote proposed different size */
 	if (size != max_payload_size) {
 		ret = firehose_send_configure(qdl, size, skip_storage_init, storage, &size);
-		if (ret < 0)
-			return ret;
+		if (ret != FIREHOSE_ACK) {
+			fprintf(stderr, "[CONFIGURE] request failed\n");
+			return -1;
```
konradybcio, line 286: might as well return ret to propagate errors from 2 levels deeper

andersson: No, because then we might return FIREHOSE_NAK from firehose_run(), I rather keep that in {0,-1} space...

PR #75, opened 2024-06-12
Indicate the progress of flashing for better user experience.

quic-vkraleti:
Add logs to firehose to show the partition label currently being flashed and progress as a percentage of completion.

### DIFF
--- a/firehose.c
+++ b/firehose.c
@@ -317,9 +317,10 @@ static int firehose_erase(struct qdl_device *qdl, struct program *program)
 	xml_setpropf(node, "num_partition_sectors", "%d", program->num_sectors);
 	xml_setpropf(node, "start_sector", "%s", program->start_sector);
 
+	fprintf(stderr, "[ERASE] Please wait, erasing \"%s\" now...\n", program->label);
 	ret = firehose_write(qdl, doc);
 	if (ret < 0) {
-		fprintf(stderr, "[PROGRAM] failed to write program command\n");
+		fprintf(stderr, "[ERASE] failed to write program command\n");
 		goto out;
 	}
 
@@ -344,6 +345,7 @@ static int firehose_program(struct qdl_device *qdl, struct program *program, int
 	void *buf;
 	time_t t0;
 	time_t t;
+	time_t tn;
 	int left;
 	int ret;
 	int n;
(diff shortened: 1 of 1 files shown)


firehose.c
```diff
@@ -401,6 +403,9 @@ static int firehose_program(struct qdl_device *qdl, struct program *program, int
 	lseek(fd, (off_t) program->file_offset * program->sector_size, SEEK_SET);
 	left = num_sectors;
 	while (left > 0) {
+		fprintf(stderr, "[PROGRAM] %d sectors remaining out of %d (%.2f%%)\r",
```
quic-bjorande, line 406: I was thinking that we didn't want to update this too often, but with the typical 1MB max_payload_size, and if my math is correct, this will generate about 10kb of logs per second on my machine. So, I don't think we need to make this more fancy at this time.

I'm thinking that we should clean up the logs a bit in general. Until then, I like this patch!

But perhaps move it last in the loop? That way it will end at 100%.

quic-vkraleti: For larger images QDL appear to be stuck without any logs. So updated the patch to limit to 1 log per second. Do check if this is fine.

firehose.c
```diff
@@ -317,9 +317,10 @@ static int firehose_erase(struct qdl_device *qdl, struct program *program)
 	xml_setpropf(node, "num_partition_sectors", "%d", program->num_sectors);
 	xml_setpropf(node, "start_sector", "%s", program->start_sector);
 
+	fprintf(stderr, "[ERASE] Please wait, erasing \"%s\" now...\n", program->label);
 	ret = firehose_write(qdl, doc);
 	if (ret < 0) {
-		fprintf(stderr, "[PROGRAM] failed to write program command\n");
+		fprintf(stderr, "[ERASE] failed to write program command\n");
```
quic-bjorande, line 323: This is a good fix, but unrelated.

firehose.c
```diff
@@ -400,7 +403,16 @@ static int firehose_program(struct qdl_device *qdl, struct program *program, int
 
 	lseek(fd, (off_t) program->file_offset * program->sector_size, SEEK_SET);
 	left = num_sectors;
+	tn = t0;
 	while (left > 0) {
+	        t = time(NULL);
```
quic-bjorande, line 408: I was expecting that you just do left % N == 0 for some value of N to avoid printing that often, but I guess this works as well...

quic-bjorande (changes requested):
I asked you to print at the end of the loop, but that obviously doesn't matter, because you have \r there, so the line will typically be overwritten.

But do we not have a problem with "flashed \"%s\" successfully" being a shorter string for short values of program->label, resulting in "corrupt" output to the log?

quic-bjorande:
Please see #88 for a more comprehensive implementation.

andersson:
Merged #88

PR #79, opened 2024-09-12
qdl: firehose: program: fail on missing file

mwasilew:
When parsing XML files the specified binary to be flashed may not be present. The default behaviour of QDL is to ignore missing file. This is sometimes undesireble.

This patch introduces optional flag --fail-on-missing which will enforce failure of flashing procedure if there is a missing file name.

### DIFF
--- a/qdl.c
+++ b/qdl.c
@@ -103,7 +103,7 @@ static void print_usage(void)
 {
 	extern const char *__progname;
 	fprintf(stderr,
-		"%s [--debug] [--storage <emmc|nand|ufs>] [--finalize-provisioning] [--include <PATH>] <prog.mbn> [<program> <patch> ...]\n",
+		"%s [--debug] [--allow-missing] [--storage <emmc|nand|ufs>] [--finalize-provisioning] [--include <PATH>] <prog.mbn> [<program> <patch> ...]\n",
 		__progname);
 }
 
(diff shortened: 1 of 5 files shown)


mwasilew:
I didn't want to change the current behaviour. The difference came out when someone reported error flashing with pcat. Pcat throws error on missing file.

quic-bjorande:
After internal discussion, I would like for us to follow pcat and invert the behavior of QDL.
Could you please make that "--allow-missing" and reverse the introduced logic?

mwasilew:
@quic-bjorande I made the changes you asked for. The code builds but I didn't test it with the device yet. I'll do that and post a comment here.

mwasilew:
It seems to be working as expected

qdl.c
```diff
@@ -55,6 +55,7 @@ enum {
 };
 
 bool qdl_debug;
+bool allow_missing;
```
quic-bjorande, line 58: This is only used as a local variable inside main(), so please move it there (and initialize it to 0).

quic-bjorande (changes requested):
One minor comment. Other than that, this looks good!
Thank you.

qdl.c
mwasilew: fixed.

PR #90, opened 2025-01-17
RFC: implement version information

### LINKED ISSUE #40: --version
It would be nice to have a `--version` on this tool, specially if it is one that uses uses the git hash when it is a development version, so we can easilly pinpoint exactly what is on the qdl binary being used. Bonus points if `--debug` by default starts by outputing the version, so qdl logs will have that info.

ndechesne:
Closes #40

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ qdl-ramdump
 ks
 compile_commands.json
 .cache
+.version.h
+version.h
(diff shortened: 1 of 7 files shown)


ljkeller:
Did a clone and build on WSL: `5.15.167.4-microsoft-standard-WSL2`

```
./qdl --version
qdl version v2.1-19-gc5ce09f
```
```
./qdl-ramdump --version
qdl-ramdump version v2.1-19-gc5ce09f
```

This change will make flashing & release management much less ambiguous for everyone working with QDL. This allows the creation of compatibility matrices with BSP releases x QDL versions.

Thank you so much 😄

obbardc (approved):

PR #92, opened 2025-01-28
Makefile: wait for versionfile before actually building

### LINKED ISSUE #91: make: *** No rule to make target 'version.h', needed by 'util.o'
Full log:

```text
cc -O2 -Wall -g `pkg-config --cflags libxml-2.0 libusb-1.0`   -c -o firehose.o firehose.c
cc -O2 -Wall -g `pkg-config --cflags libxml-2.0 libusb-1.0`   -c -o qdl.o qdl.c
cc -O2 -Wall -g `pkg-config --cflags libxml-2.0 libusb-1.0`   -c -o sahara.o sahara.c
make: *** No rule to make target 'version.h', needed by 'util.o'.  Stop.
make: *** Waiting for unfinished jobs....
```

I could easily reproduce the issue with `make -j16` but not `make -j1`.

Seems like `util.o` target may be considered before `versionfile` target finishes while using multi-threaded make. Maybe we should either make `util.o` depend on `versionfile` directly or rename `versionfile` to `version.h`?

Henry-ZHR:
Fix #91

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,9 @@ KS_OUT := ks
 KS_SRCS := ks.c sahara.c util.c ux.c
 KS_OBJS := $(KS_SRCS:.c=.o)
 
-default: versionfile $(QDL) $(RAMDUMP) $(KS_OUT)
+default: versionfile .WAIT $(QDL) $(RAMDUMP) $(KS_OUT)
+
+.WAIT:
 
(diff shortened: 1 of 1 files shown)


obbardc:
I propose an alternative fix as https://github.com/linux-msm/qdl/pull/93

PR #93, opened 2025-02-04
Some fixes for --version argument.

obbardc:
Small fixes for two errors:

- call wrong dependency
- allow VERSION to be passed into Makefile

See commit messages for details.

cc @ndechesne @andersson @quic-bjorande


Henry-ZHR:
LGTM

PR #94, opened 2025-02-21
Added Windows x64 and ARM64 support

JohnSagaQuic:
1. Added Windows specific implementation to make QDL works on Windows x64 and ARM64. 
2. Added cross platform APIs for file/directory operations
3. Added Windows building instruction
4. Added Windows specific command parameters processing


andersson:
Glancing over patches 1-3 looks reasonable, a fair amount of stylistic things, but it looks in line with my expectations for an initial patchset. I think the OS abstraction would look better if we first introduces a slightly higher level of abstraction (e.g. rather than calling fstat() let's introduce a qdl_get_file_size(), perhaps even move away from passing around raw "fd" integers)

Patch 4 on the other hand, adds 26,000 lines of (primarily) snapshots of other open-source projects. That's almost 10x the size of the QDL project. I don't see that we want to take ownership of all that code and I don't see how it can be maintained in this form.

If I understand the build steps, they call for the developer to manually build the 5 projects, then copy files to a particular place, followed by using VS 2019 to build QDL. Given that the QDL community exclusively consists of Linux and MacOS developers we absolutely need a model that will fit in Github Actions.

JohnSagaQuic:
Good suggestion on qdl_get_file_size().  Will check it further.

For the patch 4, it's only for Windows build.  What included is only header files, not source files from those open source projects.  But agreed that it's quite different experience for Linux and MacOS developers.  I also don't believe customers want to build the Windows themselves. Hence, Windows build probably only for us to generate.

EmbeddedAndroid:
> For the patch 4, it's only for Windows build. What included is only header files, not source files from those open source projects. But agreed that it's quite different experience for Linux and MacOS developers. I also don't believe customers want to build the Windows themselves. Hence, Windows build probably only for us to generate.
 
I would like to build this for Windows from source.

EmbeddedAndroid:
> Good suggestion on qdl_get_file_size(). Will check it further.
> 
> For the patch 4, it's only for Windows build. What included is only header files, not source files from those open source projects. But agreed that it's quite different experience for Linux and MacOS developers. I also don't believe customers want to build the Windows themselves. Hence, Windows build probably only for us to generate.

Might be worth looking at https://github.com/ilammy/msvc-dev-cmd to provide a Windows build env without the need for VS proj files etc.

JohnSagaQuic:
> > Good suggestion on qdl_get_file_size(). Will check it further.
> > For the patch 4, it's only for Windows build. What included is only header files, not source files from those open source projects. But agreed that it's quite different experience for Linux and MacOS developers. I also don't believe customers want to build the Windows themselves. Hence, Windows build probably only for us to generate.
> 
> Might be worth looking at https://github.com/ilammy/msvc-dev-cmd to provide a Windows build env without the need for VS proj files etc.

Bjorn, I think users will still need to install VS 2019 to compile.  And user still still need to compile  all the dependency projects (like libusb, libxml ...). We are not going to add all those source files into our repo (this PR only added the needed header files).

Changing to a different build env, will require quite some effort.  We would need someone to provide justification that there are customers really looking to build the Windows code, instead of using the Windows binary we distributed via QSC public link.  We don't want to put a lot effort and no one is using it.

quic-bjorande:
I don't doubt that most Windows users would prefer a binary, the same should be true on Linux as well. But there are certainly users who wants to help improving the project and they need to be able to build the code - and if they can't build for the different targets, we need to be able to automatically catch side effects affecting these other targets - so that a Linux patch doesn't break the build for the Windows developer, or vice versa.

I don't know how to handle dependencies in Windows, this wasn't what I was expecting. I don't think it's reasonable to turn the project into 90% header-files from 3rd party libraries. And pulling the code in as well would just grow that number...

JohnSagaQuic:
I agree with your point and concern. Pulling the source code from all the Windows dependency, will make the code size about 5 times bigger, which will make it much worse for your concern that even with just header files, already make it  "90% header-files from 3rd party libraries".

How about we do these.  
1.  Remove the Windows specified third party header files (libusb, libxml2, zlib, xz-win, libiconv-win) from the PR, which will remove that 90% concern.
2. Update the READM_WIN, with more details instruction how to build the dependencies, how to copy the needed header files into certain location.  

With these, we make common tasks easy (most users either use binary, or linux source.). We will maintain the Windows code in case change break other platform.  For corner case that some users really want to compile on Windows, they can follow the document.

ndechesne:
The plan above looks good, can you suggest these changes? Also, can you please look at adding support to build the Win binaries in .github/workflows/build.yml ?

JohnSagaQuic:
OK, I clean out the Windows folder, to remove the header files from those third party dependencies.  The size of the Windows folder is now significantly reduced. 

The READM_WIN updated with instruction to build those dependency and copy the header/libs.

For Windows, we are not using build.yml, and it's complicated to do that via commandline. Please follow the READM_WIN for build.

ndechesne:
thanks for the updates. Can you please clean up / rebase your branch? currently you have a commit that adds all the Windows dependencies/libs, and then a commit that removes them, and in fact not even all of them you still have the following left over in ./Windows:

`./unistd.h
./unistdSSize.h
./getopt.c
./sys_time_h-master
./sys_time_h-master/LICENSE
./sys_time_h-master/sys
./sys_time_h-master/sys/time.h
./getopt.h
./libiconv-win-build
./libiconv-win-build/COPYING
./libiconv-win-build/AUTHORS
./libiconv-win-build/LICENSE.md
./libiconv-win-build/THIS_VERSION_IS_1.17
./ThirdParty
./ThirdParty/BUILDING_THIRD_PARTY
./libxml2-win-build
./libxml2-win-build/THIS_VERSION_IS_2.13.5
./libusb-1.0.27
./libusb-1.0.27/COPYING
./libusb-1.0.27/AUTHORS`

Some files should clearly not be there.

ndechesne:
Also, please try to follow the convention used in the repo for the commit description and messages. Also add signed-off-by tag to your changes.

JohnSagaQuic:
1. Updated commit message to follow the convention.
2. For several files remains in the Windows folder, I kept those on purpose for reasons below.
    . Modified from third party open source files. User won't be able to directly found online
    . License related files
    . Kept the folder strcture, to make it easy for user to follow the README coping header files.

ndechesne:
Please make the following updates:
1. each patch needs a proper commit summary and description. The summary should be short (one line). The description is potentially optional for trivial commit only.
2. each patch should have a signed-off tag
3. you need to rebase/squash patch 4 and 6. We cannot add and remove things in the same PR. The commit log is expected to be a clean log. 
4. You cannot keep a bunch of  random COPYING and AUTHORS files in ./Windows folder. 

It's difficult for us to review a PR when it's not in a clean state. Once things are cleaned, we will start a review of the actual content.

I expect that not being to build for Windows without VSCode to be quite problematic. Building this branch on Linux shows a few new compiler warnings, please fix them too.

JohnSagaQuic:
1. Removed the Windows third party legal/license and version information text file.
2. Fixed the Linux compiling warnings.
3. I am not clear on what you mean the "each patch should have a signed-off tag".  Please clarify.
4. The commit message were already updated. Please let me know if any particular commit message you have concern. 
5. Not clear on what you mean "you need to rebase/squash patch 4 and 6. We cannot add and remove things in the same PR. The commit log is expected to be a clean log."   This PR includes multiple commits,  I deleted the Windows specific headers files per code review comments.

JohnSagaQuic:
I will do a clean up and probably create a totally new branch/PR to simplify the PR review.

ndechesne:
> I will do a clean up and probably create a totally new branch/PR to simplify the PR review.

Thanks for doing the cleanup! But please, do not create a new PR. If you do that, you are loosing the entire history of the review. It is a bad practice. Instead, you should clean up your branch in your local tree, and 'force push' your branch to the same branch you used when you opened the PR. It will do the right thing: 1) update the PR, 2) notify us.

ks.c
```diff
@@ -23,21 +23,20 @@ bool qdl_debug;
...
 
 int qdl_write(struct qdl_device *qdl, const void *buf, size_t len)
 {
 
-	return write(qdl->fd, buf, len);
+	return WRITE(qdl->fd, buf, len);
 }
 
-static void print_usage(void)
+static void print_usage(char * progName)
```
lumag, line 35: No CamelCase, please

ks.c
```diff
@@ -2,17 +2,17 @@
 #include <assert.h>
 #include <ctype.h>
 #include <dirent.h>
-#include <err.h>
+#include "qdl_oscompat.h"
 #include <errno.h>
 #include <fcntl.h>
 #include <getopt.h>
-#include <poll.h>
+//#include <poll.h>
```
lumag, line 9: Drop extra comments, they are present in the git history if anybody needs them

ks.c
```diff
@@ -23,21 +23,20 @@ bool qdl_debug;
 
 int qdl_read(struct qdl_device *qdl, void *buf, size_t len, unsigned int timeout)
 {
-	return read(qdl->fd, buf, len);
+	return READ(qdl->fd, buf, len);
```
lumag, line 26: I think qdl_read / qdl_write provide a nice abstraction already. Use high-level functions for OS abstraction. Another option might be to use MSYS or Cygwin in order to provide a normal libc interface. I don't understand why we need to abstract POSIX API.

qdl_oscompat.h
```diff
@@ -0,0 +1,113 @@
...
+    #define WARN printf
+    #define FSTAT64 _fstat64
+    #define STAT64 _stat64
+    #define O_BINARY _O_BINARY
+    #define O_RDONLY _O_RDONLY
+    #define O_RDWR _O_RDWR
+    #define O_WRONLY _O_WRONLY
+    #define O_CREAT _O_CREAT
+    #define O_TRUNC _O_TRUNC
+    #define PATH_MAX 256
```
lumag, line 38: Just no. Windows headers shold provide ISO C compatibility. If they don't, then fix it inside the compat layer instead of providing defines and renaming standard functions in the code.

qdl_oscompat.h
```diff
@@ -0,0 +1,113 @@
...
+	{
+		const char* name;
+		//has_arg can't be an enum because some compilers complain about
+			//type mismatches in all the code that assumes it is an int.  
+		int has_arg;
+		int* flag;
+		int val;
+	};*/
+#else
+    // Linux-specific includes
```
lumag, line 51: No, they are not Linux-specific. It's called POSIX for most of the files here.

qdl_oscompat.h
```diff
@@ -0,0 +1,113 @@
...
+
+    char** ListFiles(const char* pattern, int* count);
+
+#else
+//Linux and Mac implementation
+int OPENAT(const char* ramdump_path, const char* fileName, int ramdump_dir);
+    void CLOSEHANDLE(int fd);
+    int PATHMATCH(const char* pattern, const char* filename);
+    int OPENDIRECTORY(const char* ramdump_path);
+    void CLOSERAMDUMPHANDLE(int ramdump_dir);
```
lumag, line 110: Still, just no. Leave HANDLE to the Windows world, everything else has file descriptors.

lumag (changes requested):

ks.c
JohnSagaQuic: Got your point about POSIX. While Windows does support POSIX in general, the read() API is not natively supported due to historical and architectural reason. We tried to define this so it's straightforward for multi OS, while workaround Microsoft limit support on POSIX.

qdl_oscompat.h
JohnSagaQuic: Same reason above, Microsoft's support on POSIX is limit on Windows.

qdl_oscompat.h
JohnSagaQuic: The purpose of this change, is to keep the #if Windows stuff limit to minimal file like qdl_oscompat.h, so most the .c file doesn't need to add the platform specific check/implementation.

JohnSagaQuic:
> > I will do a clean up and probably create a totally new branch/PR to simplify the PR review.
> 
> Thanks for doing the cleanup! But please, do not create a new PR. If you do that, you are loosing the entire history of the review. It is a bad practice. Instead, you should clean up your branch in your local tree, and 'force push' your branch to the same branch you used when you opened the PR. It will do the right thing: 1) update the PR, 2) notify us.

Sent you email offline to explain a new proposal. That's why I mentioned to have a new PR with all clean up.

qdl_oscompat.h
lumag: So... My suggestion would be to come with the POSIX compatibility defines instead of the OS-compat being non-native for both OS worlds.

ks.c
JohnSagaQuic: sahara.c(217,6): **error** C4996: 'read': The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _read. See online help for details.

JohnSagaQuic: you are right, using _CRT_NONSTDC_NO_WARNINGS can suppress the warning and read() can work. Thanks for the info.

lumag:
Closed in favour of #105

PR #96, opened 2025-03-19
make: don't rebuild util.o (and dependencies) unnecessary

lumag:
Using fake target name forces make to rebuild all targets that depend on it. Properly specify version.h as a target name and use double-colon to let make know that it's a special build target.


lumag:
cc @obbardc

obbardc (approved):

obbardc:
Looks good to me.

ndechesne:
Looks good to me.. looks like the double column is the trick i should have known about..

PR #100, opened 2025-04-11
ks: Add support for libusb communication

lixkel:
This patch adds support for libusb communication with devices in edl mode, while retaining the existing implementation using file descriptors. It's unclear to me if the file descriptor implementation is actively used by anyone (it doesn't work very well for me), so feedback on its usage and potential deletion is welcome.


lumag:
cc @quic-jhugo 
Could you please take a look

quic-jhugo:
(commenting before I actually have a look at the code)
The file descriptor thing is probably not used with the upstream kernel.  I don't think the wwan subsystem has any Sahara devices exposed to userspace, although it could.
AIC100 has an in-kernel Sahara implementation now, mostly as a result of failing to get an acceptable file descriptor interface designed.
Most of the downstream Sahara implementations on Linux, and also the windows implementations, directly pipe the Sahara channel to userspace via a chardev.  This qdl project would be useful there with the file descriptor interface.

ks.c
```diff
@@ -62,6 +52,7 @@ int main(int argc, char **argv)
 		{"version", no_argument, 0, 'v'},
 		{"port", required_argument, 0, 'p'},
```
quic-jhugo, line 53: This is still listed as required, but the documentation changes above suggest that this PR is making it optional

ks.c
```diff
@@ -21,28 +21,17 @@ static struct qdl_device qdl;
...
-int qdl_read(struct qdl_device *qdl, void *buf, size_t len, unsigned int timeout)
-{
-	return read(qdl->fd, buf, len);
-}
-
-int qdl_write(struct qdl_device *qdl, const void *buf, size_t len)
-{
-
-	return write(qdl->fd, buf, len);
-}
```
quic-jhugo, line 33: How does removing these definitions not cause a compile error?

ks.c
lixkel: Thats the `has_args` field, which states that the flag requires argument not that the flag itself is required

ks.c
lixkel: These functions are defined in `qdl.h` and already implemented in `usb.c`, I edited the `usb.c` implementation so that if the `fd` field in `qdl` struct is defined the file descriptor is used instead of libusb. This solution is not ideal, but its the best I could come up with without more extensive changes to the codebase, which seem pointless as only `ks` uses file descriptors.

usb.c
```diff
@@ -217,6 +217,9 @@ int qdl_read(struct qdl_device *qdl, void *buf, size_t len, unsigned int timeout
 	int actual;
 	int ret;
 
+	if (qdl->fd != 0)
+		return read(qdl->fd, buf, len);
```
quic-kdybcio, line 221: I think defining a separate simple `chardev` backend would make more sense.. this works, but is not pretty

PR #101, opened 2025-04-16
Allow erasing all storage types

monkbroc:
Erase operations are currently only supported on devices using NAND storage. With this change, erase operations also work on devices with UFS storage.


lumag (changes requested):
Please update commit description to include DCO (Sign-off) and a suitable commit message (pull request has good enough decription, you can c&p it).

monkbroc:
Done

lumag (approved):

PR #102, opened 2025-04-22
sahara: fix ignoring return value warning

igoropaniuk:
Fix build warning in sahara protocol implementation: sahara.c: In function ‘sahara_debug64_one’:
sahara.c:348:25: warning: ignoring return value of ‘write’
      declared with attribute ‘warn_unused_result’ [-Wunused-result]
  348 |                         write(fd, buf, n);
      |                         ^~~~~~~~~~~~~~~~~

### DIFF
--- a/sahara.c
+++ b/sahara.c
@@ -306,8 +306,9 @@ static ssize_t sahara_debug64_one(struct qdl_device *qdl,
 {
 	struct sahara_pkt read_req;
 	uint64_t remain;
-	size_t offset;
+	size_t offset, buf_offset;
 	size_t chunk;
+	size_t written;
 	ssize_t n;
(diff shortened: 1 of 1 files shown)


sahara.c
```diff
@@ -345,7 +345,10 @@ static ssize_t sahara_debug64_one(struct qdl_device *qdl,
 				goto out;
 			}
 
-			write(fd, buf, n);
+			if (n != write(fd, buf, n)) {
```
lumag, line 348: I think it should be slightly different:
- error if returned value is less or equal to zero
- increase offset by the return value otherwise

igoropaniuk: thanks, fixed!

PR #103, opened 2025-04-22
qdl: introduce allow-fusing parameter 

igoropaniuk:
Introduce the `--allow-fusing` parameter, which must be explicitly set if the `secdata` partition is programmed, as it will lead to irreversible changes (fuses will be blown during the next boot).

### DIFF
--- a/program.c
+++ b/program.c
@@ -278,6 +278,27 @@ int program_find_bootable_partition(bool *multiple_found)
 	return part;
 }
 
+/**
+ * program_is_sec_partition_flashed() - find if secdata partition is flashed
+ *
+ * Returns true if filename for secdata is set in program*.xml,
+ * or false otherwise.
+ */
+int program_is_sec_partition_flashed(void)
+{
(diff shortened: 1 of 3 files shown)


igoropaniuk:
@konradybcio @lumag @andersson could you please take a look. thanks!

program.c
```diff
@@ -219,6 +219,23 @@ int erase_execute(struct qdl_device *qdl, int (*apply)(struct qdl_device *qdl, s
 	return 0;
 }
 
+static struct program *program_find_partition(const char *partition)
```
lumag, line 222: Please write somewhat sensible commit message, descring why do you need to perform these changes and/or what kind of refactoring was performed

lumag (changes requested):
LGTM except for a single comment.

program.c
igoropaniuk: Extended commit message. Let me know if you have any additional suggestions or objections. Thanks!

PR #113, opened 2025-06-16
tests: add initial test set

igoropaniuk:
Add tests make target and a simple test that executes qdl in dry-run mode
for a synthetic reference FLAT build with VIP table generation.

The FLAT build contains:
- patch0.xml
- patch1.xml
- rawprogram0.xml
- rawprogram1.xml

All binaries that these XML files point to are filled with zeros, generated
during github action  execution.

Tests ensures that the table is generated correctly by comparing
calculated and expected SHA256 hashes of DigestToSign.bin file.


.github/workflows/build.yml
```diff
@@ -62,6 +84,28 @@ jobs:
...
+          if [ "$actual_digest" = "$expected_digest" ]; then
+            echo "VIP tables are generated successfully and validated"
+            exit 0
+          else
+            echo "Expected SHA256 digest of DigestsToSign.bin file is $expected_digest"
+            echo "Calculated SHA256 digest of DigestsToSign.bin file is $actual_digest"
+            echo "VIP table folder contents:"
+            ls -la .github/dry-run-test/vip
+            exit 1
+          fi
```
lumag, line 107: Could you please extract this to a script, removing unnecessary duplication between workflow stages.

lumag (approved):

ndechesne:
Why did you add the 'test' files in .github, instead of making them part of the project, like a 'tests' folder which could be exercised with 'make tests'?

igoropaniuk:
@ndechesne  actually that's a great idea, will take a look

Makefile
```diff
@@ -42,7 +42,13 @@ clean:
 	rm -f $(KS_OUT) $(KS_OBJS)
 	rm -f compile_commands.json
 	rm -f version.h .version.h
+	rm -f tests/data/*.bin tests/data/*.img tests/data/*.elf
+	rm -rf tests/data/vip
```
lumag, line 46: Please move to the test script itself

lumag (changes requested):

igoropaniuk:
@ndechesne @lumag reworked, please take a look

lumag (approved):

quic-kdybcio (approved):

PR #146, opened 2025-11-03
For linux msm/spinor

andersson:


firehose.c
```diff
@@ -568,7 +577,7 @@ static int firehose_program(struct qdl_device *qdl, struct program *program, int
 
 			vip_transfer_clear_status(qdl);
 		}
-		n = qdl_write(qdl, buf, chunk_size * sector_size, 30000);
+		n = qdl_write(qdl, buf, chunk_size * sector_size, zlp_timeout);
```
igoropaniuk: The actual increase to 30s occurred in the previous commit `"qdl: Be explicit about storage types"`. Could you rephrase the commit message to more accurately reflect the changes made in that commit, or transfer that change (`s/10000/30000/g`) to this commit?

firehose.c
```diff
@@ -437,10 +437,12 @@ static int firehose_erase(struct qdl_device *qdl, struct program *program)
 	return ret == FIREHOSE_ACK ? 0 : -1;
 }
 
-static int firehose_program(struct qdl_device *qdl, struct program *program, int fd)
+static int firehose_program(struct qdl_device *qdl, struct program *program,
+			    int fd, enum qdl_storage_type storage)
```
igoropaniuk: Considering that `struct qdl_device` already has some storage-related fields like `size_t sector_size`, why not include `qdl_storage_type` into this structure as well?

This will remove the need pass one more param to functions like this

igoropaniuk (approved):
There is still a typo in the commit message([firehose: Increate program timeout for SPINOR](https://github.com/linux-msm/qdl/pull/146/commits/160bc76a99bf0e90308d989a3e60b6e57e87f8e4)), but besides this LGTM

PR #148, opened 2025-11-04
Makefile: isolate build artifacts and use build dir

igoropaniuk:
Previously, object files, binaries, and generated headers like .version.h were created directly in the project root. This cluttered the source tree and made it difficult to maintain separate debug and release builds. It also risked file collisions when switching build types or performing parallel builds.

* Introduced a unified build directory structure: 
```bash  
  build/
    release/
      bin/   - final executables
      obj/   - object files
      gen/   - generated headers (e.g., version.h)
    debug/
      ...
```
* Added BUILD_TYPE variable (defaults to "debug") with make debug and
  make release targets.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
           sudo apt-get install -y libxml2-dev libusb-1.0-0-dev
 
       - name: Build
-        run: make
+        run: make all
 
       - name: Run tests
         run: make tests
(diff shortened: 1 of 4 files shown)


lumag:
I think it would be better to switch to something which already has a notion of separate build dirs - e.g. meson.

igoropaniuk:
@lumag https://github.com/linux-msm/qdl/pull/150

The changes require some refinement, as I haven't added the checkpatch target yet. Additionally, I need to address an issue with the VERSION option in older versions of Meson. This is why I initially added the RFC tag, just to gather some initial feedback.

PR #150, opened 2025-11-10
[RFC] meson: replace GNU Make-based build system with Meson

igoropaniuk:
Replace the existing Makefile-based build system with Meson to improve
build performance, cross-platform support, and maintainability.

Key improvements:

* Build performance: Shared source files (sahara.c, util.c, oscompat.c,
  ux.c, sim.c, io.c) are now compiled once and linked into multiple
  targets instead of being compiled separately for each executable.
  This reduces total compilation units from ~30 to ~20 and speeds up build.

  Full builds tests benchmark results below

  Before (Make): 
  ```
    $ time make
    ...
    real    0m1.668s
    user    0m1.424s
    sys     0m0.242s
  ```
  After (Meson): 
  ```
    $ time bash -c '
    meson setup build
    meson compile -C build'
    ...
    real    0m0.759s
    user    0m0.877s
    sys     0m0.237s
  ```

* Flexible build directory: Unlike Make which builds in-tree, Meson
  supports out-of-tree builds with custom build directories.

  Example workflows:
  ```
    meson setup builddir              # Default debug build
    meson setup build-release --buildtype=release
    meson setup build-asan -Db_sanitize=address
    meson setup build-arm64 --cross-file=aarch64-linux.txt
   ```

  This enables:
  - Multiple build configurations simultaneously
  - Clean source tree (no .o files polluting git status)
  - Easy comparison between debug/release builds
  - Separate builds for different architectures

* Cross-compilation: Native support for cross-compilation via Meson
  cross-files enables building for multiple architectures without
  requiring separate build runners for each target platform.

  Before (Make): Requires native ARM64 runner in GitHub Actions
  ```
    runs-on: ubuntu-24.04-arm  # Dedicated ARM64 hardware
    run: make
  ```
  After (Meson): Cross-compile from x64 host
  ```
    runs-on: ubuntu-24.04  # x64 host
    run: |
      meson setup build-arm64 --cross-file=ci/aarch64-linux.txt
      meson compile -C build-arm64
   ```
* Automatic dependency tracking: Meson automatically tracks header
  dependencies, eliminating the need for manual dependency declarations
  and preventing missed recompilation on header changes.

  Before (Make): Manual dependency tracking
    util.o: version.h  # Must be manually maintained

  After (Meson): Automatic header scanning
    No manual declarations needed
    Changing qdl.h automatically rebuilds all files that include it

  Result: Prevents subtle bugs where changing a header doesn't trigger
  necessary recompilation.

* Platform handling: Windows-specific linking (ws2_32) and platform
  detection are now handled declaratively without shell conditionals.

* Testing integration: Test suite is now integrated into the build system
  with proper dependency tracking and timeout handling.

  Before (Make):
    `make tests  # Shells out to bash script`

  After (Meson):
  ```
    meson test -C builddir
    meson test -C builddir --verbose
    meson test -C builddir --wrapper='valgrind --leak-check=full'
  ```

  Result: Better test output formatting, timeout handling, and ability
  to run tests under valgrind or sanitizers.

* Installation: Standard installation paths and packaging support through
  Meson's built-in install infrastructure.

  Before (Make):
    `make install prefix=/usr/local DESTDIR=/tmp/package`

  After (Meson):
  ```
    meson setup builddir --prefix=/usr/local
    DESTDIR=/tmp/package meson install -C builddir
    # Correctly installs to:
    # - /usr/local/bin/{qdl,qdl-ramdump,ks}
    # - /usr/local/share/man/man1/{qdl.1,qdl-ramdump.1,ks.1}
    ```

  Result: Better integration with distribution packaging (Debian .deb,
  RPM) and standard FHS paths.

Migration maintains full feature parity with the existing build system:
- All three executables (qdl, qdl-ramdump, ks) build correctly
- Version string generation from git tags
- Optional man page generation
- Support for Linux, macOS, and Windows (MSYS2/MinGW)
- Recreated tests integration via Meson’s test() function,
  including support for script arguments (e.g. --builddir)

Usage:
```
$ meson setup build
$ meson compile -C build
$ meson test -C build
$ meson compile check -C build
```


igoropaniuk:
@lumag all issues are fixed (I just need to add checkpatch target, will do this later).

andersson:
Why? What problem does switching to meson solve?

igoropaniuk:
@andersson The motivation for this change originated from another PR, https://github.com/linux-msm/qdl/pull/148, and  suggestions there to explore better solutions like Meson

z3ntu:
Also makes it possible to easily build a static qdl with `LDFLAGS="-static" meson setup builddir --prefer-static && meson compile -C builddir`

igoropaniuk:
Rebased onto the latest master

igoropaniuk:
I understand the concern that switching from Make to Meson might be seen as unnecessary overhead. I tried to outline the motivation and benefits in the commit message, but I’m happy to clarify or expand on any specific points if something is still unclear.

To help with a smooth transition, I can also prepare a compatibility wrapper that preserves the existing make, make install, and make check targets, while using Meson under the hood. That way, existing workflows would continue to work unchanged.

At this point, I’d appreciate your guidance on whether you see a realistic path toward merging this change (possibly with adjustments), or whether you feel this direction is unlikely to be accepted. I want to make sure I’m not spending time iterating on something that doesn’t align with the project’s direction.

.github/workflows/build.yml
```diff
@@ -24,23 +24,49 @@ jobs:
...
+          }
+
+          current="$(version_to_num "$meson_version_short")"
+          threshold="$(version_to_num "1.1")"
+          # According to https://mesonbuild.com/Build-options.html for versions < 1.1 we should use
+          # meson_options.txt for build options
+          if (( current < threshold )); then
+              if [[ -f "meson.options" ]]; then
+                  echo "Detected Meson $meson_version (< 1.1) - renaming meson.options -> meson_options.txt"
+                  mv -f meson.options meson_options.txt
```
lumag: Huh? Don't play with the repo, please. Stick to one of the names and use meson version checks in the source to prevent it from being built by the out-of-date meson.

PR #153, opened 2025-11-13
Suppress unused var warnings

igoropaniuk:
1. Introduce __unused macro to prevent warnings when a function parameter
    might not be used.
2. Use __unused attribute to suppress compile warnings for callback
    implementations, where some parameters aren't used


igoropaniuk:
@andersson now in the checks list  I see only `macos-13, intel`, but no `macos-15` anymore :(

PR #158, opened 2025-11-21
firehose: Add slot flag

jcreedon:
For SoCs that support multiple UFS devices, firehose allows the slot property. This adds a flag to set the slot parameter.

### DIFF
--- a/firehose.c
+++ b/firehose.c
@@ -417,6 +417,9 @@ static int firehose_erase(struct qdl_device *qdl, struct program *program)
 	xml_setpropf(node, "num_partition_sectors", "%d", program->num_sectors);
 	xml_setpropf(node, "physical_partition_number", "%d", program->partition);
 	xml_setpropf(node, "start_sector", "%s", program->start_sector);
+	if (qdl->slot != UINT_MAX) {
+		xml_setpropf(node, "slot", "%u", qdl->slot);
+	}
 	if (program->is_nand) {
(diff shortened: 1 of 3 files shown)


firehose.c
```diff
@@ -878,6 +896,9 @@ int firehose_apply_ufs_epilogue(struct qdl_device *qdl, struct ufs_epilogue *ufs
 
 	xml_setpropf(node_to_send, "LUNtoGrow", "%d", ufs->LUNtoGrow);
 	xml_setpropf(node_to_send, "commit", "%d", commit);
+	if (qdl->slot > 0) {
```
konradybcio: passing `slot="0"` is valid too

qdl.c
```diff
@@ -431,6 +431,7 @@ static void print_usage(FILE *out)
 	fprintf(out, " -S, --serial=T\t\t\tSelect target by serial number T (e.g. <0AA94EFD>)\n");
 	fprintf(out, " -u, --out-chunk-size=T\t\tOverride chunk size for transaction with T\n");
 	fprintf(out, " -t, --create-digests=T\t\tGenerate table of digests in the T folder\n");
+	fprintf(out, " -T, --slot=T\t\t\tSet slot number T for multiple UFS devices (T > 0)\n");
```
konradybcio: This doesn't only apply to UFS

konradybcio (approved):

PR #159, opened 2025-11-25
[housekeeping] Integer-related compiler warnings

igoropaniuk:


firehose.c
```diff
@@ -555,7 +556,7 @@ static int firehose_program(struct qdl_device *qdl, struct program *program, int
 				goto out;
 			}
 
-			if (n < qdl->max_payload_size)
+			if ((size_t)n < qdl->max_payload_size)
```
igoropaniuk: the check of `n < 0` is above, so safe to cast here

PR #161, opened 2025-11-28
Address memory leaks/mitigate buffer overflows

igoropaniuk:


firehose.c
```diff
@@ -390,6 +395,8 @@ static int firehose_try_configure(struct qdl_device *qdl, bool skip_storage_init
 				break;
 			}
 		}
+
+		free(buf);
```
andersson: I think replacing the malloc() with alloca() would be cleaner, then we don't have to explicitly free the memory (and we don't need the NULL check...)

qdl.c
```diff
@@ -193,44 +195,70 @@ static int decode_programmer_archive(struct sahara_image *images)
...
 		ptr += filesize;
 		ptr = ALIGN_UP(ptr, 4);
 	}
 
-	free(blob->ptr);
-	blob->ptr = NULL;
-	blob->len = 0;
+	/*
+	 * even if some images aren't initialized, it's safe to enumerate over each
+	 * image and try to free, as otherwise image[i] will be filled with zeros.
```
andersson: If needed, this would better be added to the comment above the function, stating that the images array must be zero-initialized (but I don't think it's necessary to document that).

qdl.c
```diff
@@ -193,44 +195,70 @@ static int decode_programmer_archive(struct sahara_image *images)
...
+	 *
+	 * first image isn't freed if error occurred
+	 */
+	for (size_t i = 0; i < MAPPING_SZ; i++) {
+		if (ret != 1 && i == 0)
```
andersson: Please see if you can write this in a less clever way, or perhaps add the comment here.

qdl.c
```diff
@@ -193,44 +195,70 @@ static int decode_programmer_archive(struct sahara_image *images)
...
+			continue;
 
-	return 1;
+		if (images[i].ptr)
+			free(images[i].ptr);
+		if (images[i].name)
+			free((void *)images[i].name);
```
andersson: Perhaps drop the const instead of type casting here?

igoropaniuk:
Friendly ping on this PR.
Let me know if any changes are needed. Thanks!

qdl.c
```diff
@@ -395,7 +395,11 @@ static int decode_programmer(char *s, struct sahara_image *images)
 			return -1;
 
 		ret = decode_programmer_archive(&archive, images);
-		if (ret < 0 || ret == 1)
+		if (ret < 0) {
+			sahara_images_free(images, MAPPING_SZ);
```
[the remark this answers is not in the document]
igoropaniuk: Moved sahara_images_free() inside decode_programmer_archive()

qdl.c
```diff
@@ -395,7 +395,11 @@ static int decode_programmer(char *s, struct sahara_image *images)
...
+			return ret;
+		}
+		if (ret == 1)
 			return ret;
 
 		ret = decode_sahara_config(&archive, images);
```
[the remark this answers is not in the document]
igoropaniuk: Fixed, thanks!

PR #164, opened 2025-12-09
[RFC] qdl: make timeout configurable

igoropaniuk:
Unify the timeout value (set default to 30000ms), propagate it through all architectural layers that use the same QDL structure, and make it configurable via a CLI parameter so it can be increased when needed.

### DIFF
--- a/firehose.c
+++ b/firehose.c
@@ -212,7 +212,7 @@ static int firehose_write(struct qdl_device *qdl, xmlDoc *doc)
 		return -1;
 	}
 	if (vip_transfer_status_check_needed(qdl)) {
-		ret = firehose_read(qdl, 30000, firehose_generic_parser, NULL);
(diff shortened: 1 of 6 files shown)


andersson:
I guess there's might not really be a point to optimize the failure case and keep a lower non-SPI-NOR timeout, we should after all not hit that case...

I don't like the idea of adding the -w option, and I definitely don't like the idea that one of our platforms require that you pass -w with some number. We should make sure that the tool work out of the box, rather than depend on tribal knowledge.
I've reached out to storage team to see what impacts this time and see if we can make a more educated decision (30 was chosen because it was 2x the worst case I measured).

JohnSagaQuic:
Just share some info here.  I think either using 30s timeout, or using the -w option as the code implemented here are fine.  PCAT tool use the option "Firehose data reception timeout"

PR #170, opened 2025-12-17
qdl: Allow skipping the firehose step by omitting xml files

jerry-skydio:
Sometimes it is useful to run only the sahara step without running
firehose. Allow that by detecting if no xml files are passed.

This is mainly useful to mix and match qdl with stock tools like
QSaharaServer and fh_loader, in case some combination of these
behaves better for a given device.

Signed-off-by: Jerry Zhang <Jerry@skydio.com>


andersson:
I think we need to combine this feature with the support for then running qdl again, while the device is already in Firehose state.

quic-kdybcio:
Isn't this what `ks` ("kickstart") already achieves though?

PR #175, opened 2025-12-22
firehose: Increase reset delay after flashing

ykaire-qti:
Added a 10-second reset delay after flashing to improve stability on RB3 and IQ9 platforms

Signed-off-by: Yvonne Kaire <ykaire@qti.qualcomm.com>


JohnSagaQuic:
1. The 10-second delay occurs after flashing the image and before rebooting the device. Through extensive trial and error, we discovered that without a 5–7 second delay, the device may hang and require a full power cycle (disconnect and reconnect the power supply). To be safe, we implemented a 10-second delay for Rb3, IQ9.
2. PCAT also uses the same 10-second delay for the same stability reasons

quic-bjorande:
Thank you for the updated data @JohnSagaQuic.

You answer clarifies the size of the problem, but not what the problem is. The statement is still that the change will "improve stability".

Also, the motivation needs to go in the commit message of the change - so that it's captured in the git history once merged. So @ykaire-qti please incorporate these changes - and please resolve the buildtest issues reported.

PR #176, opened 2026-01-12
usb: add --list flag for list devices 

lucarin91:
I think it could be useful to expose a command to list all EDL-compatible devices.

### DIFF
--- a/qdl.c
+++ b/qdl.c
@@ -417,6 +417,7 @@ static void print_usage(FILE *out)
 
 	fprintf(out, "Usage: %s [options] <prog.mbn> (<program-xml> | <patch-xml> | <read-xml>)...\n", __progname);
 	fprintf(out, "       %s [options] <prog.mbn> ((read | write) <address> <binary>)...\n", __progname);
+	fprintf(out, "       %s list\n", __progname);
 	fprintf(out, " -d, --debug\t\t\tPrint detailed debug info\n");
 	fprintf(out, " -v, --version\t\t\tPrint the current version and exit\n");
 	fprintf(out, " -n, --dry-run\t\t\tDry run execution, no device reading or flashing\n");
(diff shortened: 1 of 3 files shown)


qdl.c
```diff
@@ -525,6 +526,17 @@ int main(int argc, char **argv)
...
 			slot = (unsigned int)strtoul(optarg, NULL, 10);
 			break;
+		case 'L':
+			struct qdl_device_desc *qdl_list;
+			int n = usb_list(&qdl_list);
+			if (n <= 0) {
+				return 1;
+			}
+			for (int i = 0; i < n; i++) {
+				printf("Device %d:%d %s\n", qdl_list[i].idVendor, qdl_list[i].idProduct, qdl_list[i].iProduct);
```
andersson: I think it's more common to see vid/pid as 4-digit hex, so how about %04x:%04x instead of the %d:%d?

usb.c
```diff
@@ -223,6 +223,66 @@ static int usb_open(struct qdl_device *qdl, const char *serial)
...
+		}
+
+		char buf[128];
+		ret = libusb_get_string_descriptor_ascii(handle, desc.iProduct, (unsigned char *)buf, sizeof(buf));
+		if (ret < 0) {
+			warnx("failed to read iProduct descriptor: %s", libusb_strerror(ret));
+			continue;
+		}
+
+		(*devices)[i].idVendor = desc.idVendor;
```
andersson: Based on the various continue statements above, **i** will be a sparse value space and I believe what you intend to return is the dense list of devices that passed the filtering, i.e. the one indexed by **count**.

andersson (changes requested):
I really like this feature, please find some requests for improvements inline.

Rather than the "feat: " subject prefix, I think "usb: " would be better. Also please do include a commit message stating why this patch exist (to discover the serial numbers to be passed to --serial I presume?).

Please also add you Signed-off-by (git commit --amend -s) to give your certificate of the origin of this code.

PR #197, opened 2026-03-12
qdl: Add support for multiple include folders using a list

WilliamBright-IMD:
This is particularly useful for yocto builds since binaries are generated into three folders whereas currently QDL only supports searching for files in two folders (if you pass a single -i argument)

You can use this new feature like below:
`qdl -i folder1 -i folder2`

### DIFF
--- a/program.c
+++ b/program.c
@@ -148,13 +148,14 @@ static int program_load_sparse(struct program *program, int fd)
 	return 0;
 }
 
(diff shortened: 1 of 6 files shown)


WilliamBright-IMD:
@andersson is it worth me rebasing these patches? Is there anything else that will need changing before these can be considered

PR #201, opened 2026-03-24
[RFC] Support workflows where the Firehose programmer is already running

igoropaniuk:
Support two new CLI options to support workflows where the Firehose programmer
is already running on the device:

- --skip-reset (-R): suppresses the post-flash power/reset command, leaving the programmer alive after flashing completes.

- --skip-sahara (-P): skips the Sahara phase entirely and connects directly to an already-running programmer. When used, the programmer ELF argument is omitted. Intended to be paired with --skip-reset from a previous qdl invocation.

Typical use:

```bash
# First flash - uploads programmer and flashes, but doesn't reset
qdl prog_firehose_ddr.elf rawprogram*.xml patch*.xml --debug --skip-reset

# Subsequent flashes -  programmer already running, skip Sahara entirely
qdl rawprogram*.xml patch*.xml --skip-sahara
```

### DIFF
--- a/firehose.c
+++ b/firehose.c
@@ -1031,7 +1031,8 @@ int firehose_provision(struct qdl_device *qdl)
 	else
 		ux_info("UFS provisioning failed\n");
 
-	firehose_reset(qdl);
+	if (!qdl->skip_reset)
+		firehose_reset(qdl);
 
 	return ret;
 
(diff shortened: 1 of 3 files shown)


igoropaniuk:
Closing this for now, since it needs a more sophisticated implementation, as discussed earlier.

With this patch, we also confirmed that multiple storage types can be flashed within the same firehose_programmer session (configure firehose cmd can be send multiple times).

PR #207, opened 2026-04-04
[RFC] Introduce unit tests

igoropaniuk:
Introduce test suite for the patch module.

The library had no unit tests because there was no way to link just the protocol logic without pulling in main(), no way to silence its output, and no way to reset per-module state between test cases. 

All three of those blockers are removed here: 
- sources are collected into a static archive so a test binary can link them without a CLI entry point
- output is routed through a pluggable vtable so tests can install a no-op backend
- the patch module's file-scoped globals are moved into struct qdl_device so each test case gets a clean instance.

With those foundations in place the first real test suite is added for the patch module, and CI is updated to install cmocka so the tests run on every platform automatically.


igoropaniuk:
Closing this PR for now; the feature will be landed later across multiple PRs.

PR #209, opened 2026-04-06
Decouple the library

igoropaniuk:
Decouple the library so it can be linked independently, silence noisy side effects, and make I/O pluggable - all prerequisites for running unit tests without a device or terminal


igoropaniuk:
@andersson @lumag please take a look

PR #210, opened 2026-04-09
For linux msm/flashmap support

andersson:
When distributing builds it's cumbersome and error prone to rely on the user to run QDL on the right programmer, program, and patch files, for each memory type. In Qualcomm's Windows builds this is managed by packaging all the relevant files into a zip archive and include a "flashmap" definition, which tells the flash tools how to process the files across the different memories.

In order to implement this mechanism in QDL the various global lists of firehose operations are merged to a single list, where the client can "mix" configure, program, patch, read, etc operations. Support for parsing the current version of the flashmap json file is introduced, followed by support for getting all these files directly out of a zip file.

The end result is the ability to completely flash something like Hamoa CRD by issuing:

  qdl flash installer.zip

Worth pointing out is that the current version of the flashmap json format does not support specifying multiple programmers, so recent devices such as Nord are not yet supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ IDs `9008`, `900e`, `901d`) to upload a flash loader and use it to flash images.
 ### Linux
 
 ```bash
-sudo apt install libxml2-dev libusb-1.0-0-dev meson ninja-build help2man
+sudo apt install libxml2-dev libusb-1.0-0-dev libzip-dev meson ninja-build help2man
 meson setup build
 meson compile -C build
 ```
@@ -21,7 +21,7 @@ meson compile -C build
 For Homebrew users:
 
 ```bash
-brew install libxml2 libusb meson ninja help2man
+brew install libxml2 libusb libzip meson ninja help2man
 meson setup build
 meson compile -C build
 ```
@@ -29,7 +29,7 @@ meson compile -C build
 For MacPorts users:
 
 ```bash
-sudo port install libxml2 libusb meson ninja help2man
+sudo port install libxml2 libusb libzip meson ninja help2man
 meson setup build
 meson compile -C build
 ```
@@ -49,6 +49,7 @@ pacman -S mingw-w64-x86_64-meson
 pacman -S mingw-w64-x86_64-ninja
 pacman -S mingw-w64-x86_64-libusb
 pacman -S mingw-w64-x86_64-libxml2
+pacman -S mingw-w64-x86_64-libzip
 ```
 
 Then use the `meson` tool to build QDL:
--- a/README.md
+++ b/README.md
@@ -88,6 +88,26 @@ the board to flash through the `--serial` option:
 qdl --serial=0AA94EFD prog_firehose_ddr.elf rawprogram*.xml patch*.xml
 ```
 
+### Flashing installer packages
+
+If you have an installer package instead of individual binaries and XML
+definitions, you can flash this using the *flash* subcommand:
+
+```bash
+qdl flash <installer.zip>
+```
+
+If the *installer package* is unpacked it can be installed as:
+
+```bash
+qdl flash flashmap.json
+```
+
+These can of course be combined with e.g. *--serial*.
+
+A subset of the installer package can be selected for installation by appending
+a **::storage1[,storage2...]** suffix to the file name.
+
 ### Flash simulation (dry run)
 
 Use the `--dry-run` option to run QDL without connecting to or flashing any
--- a/file.c
+++ b/file.c
@@ -5,34 +5,71 @@
 #include <fcntl.h>
 #include <stdlib.h>
 #include <unistd.h>
+#include <zip.h>
 
 #include "oscompat.h"
 #include "qdl.h"
 #include "file.h"
 
-int qdl_file_open(const char *filename, struct qdl_file *file)
+struct qdl_zip {
+	zip_t *zip;
+	unsigned int refcount;
+};
+
+int qdl_file_open(struct qdl_zip *qdl_zip, const char *filename, struct qdl_file *file)
 {
+	struct zip_stat st;
+	zip_int64_t idx;
+	zip_file_t *zf;
 	off_t len;
+	zip_t *zip;
 	int fd;
 
-	fd = open(filename, O_RDONLY | O_BINARY);
-	if (fd < 0) {
-		ux_err("failed to open \"%s\" for reading\n", filename);
-		return -1;
+	if (qdl_zip) {
+		zip = qdl_zip->zip;
+
+		idx = zip_name_locate(zip, filename, 0);
+		if (idx < 0) {
+			ux_err("unable to locate \"%s\" in zip archive\n", filename);
+			return -1;
+		}
+
+		if (zip_stat_index(zip, idx, 0, &st) < 0) {
+			ux_err("unable to stat \"%s\" in zip archive\n", filename);
+			return -1;
+		}
+
+		zf = zip_fopen_index(zip, idx, 0);
+		if (!zf) {
+			ux_err("unable to open \"%s\" in zip archive\n", filename);
+			return -1;
+		}
+
+		file->type = QDL_FILE_TYPE_ZIP;
+		file->fd = -1;
+		file->size = st.size;
+		file->zip_file = zf;
+	} else {
+		fd = open(filename, O_RDONLY | O_BINARY);
+		if (fd < 0) {
+			ux_err("failed to open \"%s\" for reading\n", filename);
+			return -1;
+		}
+
+		len = lseek(fd, 0, SEEK_END);
+		if (len < 0) {
+			ux_err("failed to find end of \"%s\"\n", filename);
+			close(fd);
+			return -1;
+		}
+
+		lseek(fd, 0, SEEK_SET);
+
+		file->type = QDL_FILE_TYPE_POSIX;
+		file->fd = fd;
+		file->size = len;
+		file->zip_file = NULL;
 	}
-
-	len = lseek(fd, 0, SEEK_END);
-	if (len < 0) {
-		ux_err("failed to find end of \"%s\"\n", filename);
-		close(fd);
-		return -1;
-	}
-
-	lseek(fd, 0, SEEK_SET);
-
-	file->type = QDL_FILE_TYPE_POSIX;
-	file->fd = fd;
-	file->size = len;
 
 	return 0;
 }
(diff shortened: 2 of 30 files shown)


file.c
```diff
@@ -0,0 +1,211 @@
...
+	} else {
+		fd = open(filename, O_RDONLY | O_BINARY);
+		if (fd < 0) {
+			ux_err("failed to open \"%s\" for reading\n");
+			return -1;
+		}
+
+		len = lseek(fd, 0, SEEK_END);
+		if (len < 0) {
+			ux_err("fialed to find end of \"%s\"\n", filename);
```
igoropaniuk: failed

file.c
```diff
@@ -0,0 +1,211 @@
...
+			close(fd);
+		}
```
igoropaniuk, line 64: missing return on lseek failure

flashmap.c
```diff
@@ -0,0 +1,278 @@
...
+
+	ret = qdl_file_open(zip, zip ? "flashmap.json" : filename, &flashmap);
+	if (ret < 0) {
+		ux_err("failed to open flashmap\n");
+		goto out_put_zip;
+	}
+
+	json_blob = qdl_file_load(&flashmap, &json_size);
+	qdl_file_close(&flashmap);
+	if (!json_blob)
```
igoropaniuk, line 242: `ret` will be 0 here. Same issue for the `!json`, `!version || strcmp(version, "1.1.0")` - several goto out_free_json paths don't set ret = -1

flashmap.c
```diff
@@ -0,0 +1,278 @@
...
+	int ret;
+
+	ret = qdl_file_open(zip, filename, &file);
+	if (ret < 0) {
+		ux_err("unable to parse XML: %s\n", filename);
+		return -1;
+	}
+
+	xml = qdl_file_load(&file, &len);
+	if (len > INT_MAX) {
```
igoropaniuk, line 81: `len` may be uninitialized if `xml == NULL`, need to check xml as well

flashmap.c
```diff
@@ -0,0 +1,278 @@
...
+		ux_err("XML too large: %s\n", filename);
+		free(xml);
+		return -1;
+	}
+
+	doc = xmlReadMemory(xml, (int)len, filename, NULL, 0);
+
+	root = xmlDocGetRootElement(doc);
```
igoropaniuk, line 94: no NULL check on doc

firehose.c
```diff
@@ -1038,53 +1075,118 @@ int firehose_provision(struct qdl_device *qdl)
...
 
-	ret = firehose_detect_and_configure(qdl, false, qdl->storage_type, 5);
-	if (ret)
-		return ret;
+			ret = gpt_resolve_deferrals(qdl, ops);
+			if (ret)
+				return ret;
+
+			/* Update the number of patches for this storage device */
+			patch_count = 0;
```
igoropaniuk, line 1129: `patch_count` resets per configure block, but `patch_idx` is never reset. On multi-storage flashmaps the second progress run will start at e.g. `patch_idx=17 of 4` and status_patch is overwritten - likely the "N patches applied" info line will fire on the wrong patch.

qdl.h
```diff
@@ -10,6 +10,7 @@
 #endif
 
 #include <stdbool.h>
+#include <zip.h>
```
igoropaniuk: do we need it here as well?

file.h
```diff
@@ -0,0 +1,39 @@
...
+void qdl_file_close(struct qdl_file *file);
+size_t qdl_file_getsize(struct qdl_file *file);
+ssize_t qdl_file_read(struct qdl_file *file, void *buf, size_t len);
+off_t qdl_file_seek(struct qdl_file *file, off_t offset, int whence);
+
+int qdl_zip_open(const char *filename, struct qdl_zip **__qdl_zip);
+struct qdl_zip *qdl_zip_get(struct qdl_zip *qzip);
+void qdl_zip_put(struct qdl_zip *qzip);
+#endif
+
```
igoropaniuk: trailing blank line

README.md
```diff
@@ -87,6 +88,26 @@ the board to flash through the `--serial` option:
 qdl --serial=0AA94EFD prog_firehose_ddr.elf rawprogram*.xml patch*.xml
 ```
 
+### Flashing installer packages
+
+If you have an installer package intead of individual binaries and XML
```
igoropaniuk: instead

json.c
```diff
@@ -0,0 +1,441 @@
...
+	return 1;
+}
+
+static int json_parse_number(struct json_value *value)
+{
+	char buf[20];
+	char *b = buf;
+	int ch;
+
+	while ((ch = input()) && isdigit(ch) && (size_t)(b - buf) < sizeof(buf) - 1)
```
igoropaniuk: isdigit() doesn't support no negatives, decimals etc, so strtod() on line 110 might be just an overkill

`strtol()` should be sufficient

json.c
```diff
@@ -0,0 +1,441 @@
...
+	}
+
+	return NULL;
+}
+
+void json_free(struct json_value *value)
+{
+	struct json_value *next;
+	struct json_value *it;
+
```
igoropaniuk, line 816: we should check value == NULL before `free(value->key)`

list.h
```diff
@@ -8,6 +8,12 @@
 #include <stdbool.h>
 #include <stddef.h>
 
+#ifndef container_of
+#define container_of(ptr, typecast, member) ({                  \
```
igoropaniuk: we have already definition for that in `qdl.h`, please avoid duplication

igoropaniuk (changes requested):

igoropaniuk:
@andersson Could you please also add an integration dry-run test with a minimal zip file to the PR?

flashmap.c
andersson: I'm also leaking "file".

qdl.h
andersson: Leftover since before I wrapped the zip_file_t inside qdl_file. Found in 3 more files... Thanks

json.c
andersson: I did go a bit overboard and introduced a unit test for the json module, and instead of reducing this I ended up implementing this and a few other things...

andersson:
> @andersson Could you please also add an integration dry-run test with a minimal zip file to the PR?

Added both a dry-run test of flashmap.json and flashmap.zip, and then two specific unit-tests for json and flashmap parsing to ensure those are parsed as expected.

andersson:
Dropped the two unit tests from the PR for now, while I'm trying to make them run on Windows and get past the checkpatch warnings.

The two commits with tests can be found at https://github.com/quic-bjorande/qdl/commits/for-linux-msm/flashmap-support-test/

@igoropaniuk I would love some input on the latter.

igoropaniuk:
@andersson I've tried to troubleshoot the issue you mentioned and my assumption why flashmap_zip test likely fails on Windows is that `/tmp` is passed straight to Win32 API (the path is hardcoded in both `tests/test_flashmap_zip.c:777` and `tests/test_flashmap_zip.c:808`), it then resolves against the current drive (`D:` per the log), and `D:\tmp` doesn't exist on the runner - so `mkdtemp() `likely fails  with `ENOENT` before any fixture is created and every subtest bails out (I don't have access to the final log, so it's rather assumption than cofirmed fact).

The previous json test passes because it never touches the filesystem - no `mkdtemp()` (and no `/tmp`)

On the Windows job in `.github/workflows/build.yml` the test binary is built with MinGW-w64 (mingw-w64-x86_64-gcc) and therefore links against the native Windows/MSVCRT runtime - not MSYS2's POSIX emulation (native Windows CRT, so no MSYS2 path translation).

igoropaniuk:
btw, can we consider using an existing unit test framework rather than building our own test infrastructure (with our own macro for asserts, fixtures etc)? For example, `cmocka` (https://cmocka.org/) provides per-test setup/teardown, portable fixture helpers, and syscall mocking, so problems like a missing /tmp on Windows are reported as explicit test failures or skips instead of a silent exit status 1.

Example of usage of cmocka was in https://github.com/linux-msm/qdl/pull/207

qdl.c
```diff
@@ -568,10 +573,116 @@ static int qdl_ramdump(int argc, char **argv)
...
+	char *save;
+	char *tmp;
+	int ret;
+
+	tmp = strstr(args, "::");
+	if (tmp) {
+		*tmp = '\0';
+		filter = tmp + 2;
+
+		for (tmp = strtok_r(filter, ",", &save); tmp; tmp = strtok_r(NULL, ",", &save)) {
```
igoropaniuk, line 624: TBH, I don't like the idea of mutating `args` here. 

If we decide in future to print/reuse the argv somewhere after this call (error messages etc), we will find out that it's been silently truncated

util.c
```diff
@@ -221,49 +222,33 @@ int parse_storage_address(const char *address, int *physical_partition,
...
+	if (ret < 0) {
 		ux_err("failed to read \"%s\"\n", filename);
 		return -1;
 	}
 
-	len = lseek(fd, 0, SEEK_END);
-	if (len < 0) {
-		ux_err("failed to find end of \"%s\"\n", filename);
+	ptr = qdl_file_load(&file, &len);
+	if (!ptr)
```
igoropaniuk: I think we should also add a check for the empty file as well (`len == 0`), as later we unconditionally alloc `malloc(file->size)`.

file.c
```diff
@@ -0,0 +1,212 @@
...
+{
+	switch (file->type) {
+	case QDL_FILE_TYPE_UNKNOWN:
+		break;
+	case QDL_FILE_TYPE_POSIX:
+		close(file->fd);
+		file->fd = -1;
+		break;
+	case QDL_FILE_TYPE_ZIP:
+		zip_fclose(file->zip_file);
```
igoropaniuk, line 127: `file->zip_file = NULL;` after?

flashmap.c
```diff
@@ -0,0 +1,311 @@
...
+
+	ret = qdl_file_open(zip, filename, &file);
+	if (ret < 0) {
+		ux_err("unable to parse XML: %s\n", filename);
+		return -1;
+	}
+
+	xml = qdl_file_load(&file, &len);
+	if (!xml) {
+		ux_err("failed to \"%s\"\n", filename);
```
igoropaniuk: `failed to load`

igoropaniuk (commented):
LGTM! just some minor comments

PR #213, opened 2026-04-20
usb: Allow devices with no serial number to show in the usb list

Zephyrus29:
I have a module type that isn't reporting a serial number (msm8937 based). 
Currently they can't load at all, this change means while they still cannot be selected by serial number, they can still load.

### DIFF
--- a/usb.c
+++ b/usb.c
@@ -367,21 +367,19 @@ struct qdl_device_desc *usb_list(unsigned int *devices_found)
 
 		serial = strstr((char *)buf, "_SN:");
 		if (!serial) {
-			ux_err("ignoring device with no serial number\n");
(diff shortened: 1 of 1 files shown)


andersson:
I think your patch looks good, but your commit subject talks about "loading" and your patch affect the listing of devices.

Please clarify the intent with your patch, by describing the problem you're seeing in the commit message.
Perhaps as this does relate to listing it would be more useful to list the item as "(none)" or something? So that it shows up in the output?

Zephyrus29:
I can change it clarify that it affects the listing and I can change it so the serial is (none). 
Currently they show up as a VID:PID but the serial is an empty string.

PR #214, opened 2026-04-20
misc: changed file loading so that all files are searched for in the include directory

Zephyrus29:
This changes the file path search logic everywhere so that it looks for every file in the include dir

This changes someone calling it like this 
qdl -i foo foo/prog_firehose_ddr.elf foo/rawprogram*.xml foo/patch*.xml

to 

qdl -i foo prog_firehose_ddr.elf rawprogram*.xml patch*.xml

### DIFF
--- a/ks.c
+++ b/ks.c
@@ -103,7 +103,7 @@ int main(int argc, char **argv)
 				return 1;
 			}
 			filename = &optarg[colon - optarg + 1];
-			ret = load_sahara_image(filename, &mappings[file_id]);
+			ret = load_sahara_image(filename, &mappings[file_id], NULL);
 			if (ret < 0)
 				return 1;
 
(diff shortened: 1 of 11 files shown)


quic-bjorande:
Thank you for your PR, @Zephyrus29.

I typically just step into "foo" and run the command from there, so this problem hasn't really dawned on me. But I agree the current scheme is rather unintuitive.

I did play around with this a bit; one thing I don't like with your proposed form is that I can't tab-complete any of the filenames. I need to type them out in full, relative to the path I provided through -i.

The result of this ended up being #223 where I propose we look for files adjacent to the rawprogramN.xml file that referenced the file (with -i as a higher priority override option).

Your proposed command becomes:
`qdl foo/prog_firehose_ddr.elf foo/rawprogram*.xml foo/patch*.xml`

This removes the unintuitive requirement of specifying -i when you already told the tool where to find the thing you want to flash.

PR #218, opened 2026-04-20
usb: add PID 0x90db and refactor PID check

mukeshojha-linux:
Add PID 0x90db used by some Qualcomm targets in EDL mode.  Also refactor usb_try_open() to use usb_is_edl_pid() instead of a duplicate hardcoded PID list, so future PID additions only need a single change.

### DIFF
--- a/usb.c
+++ b/usb.c
@@ -36,6 +36,8 @@ struct qdl_device_usb {
 #define LIBUSB_ENDPOINT_TRANSFER_TYPE_BULK LIBUSB_TRANSFER_TYPE_BULK
 #endif
 
+static bool usb_is_edl_pid(uint16_t pid);
+
 static bool usb_read_serial(struct libusb_device_handle *handle,
 			    const struct libusb_device_descriptor *desc,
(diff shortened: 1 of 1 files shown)


andersson:
It used to be that this list gave us "definitive" answer that the device is in EDL mode, but if I read correctly this isn't a PID reserved for EDL mode. Is there some other check we should include here?

mukeshojha-linux:
> It used to be that this list gave us "definitive" answer that the device is in EDL mode, but if I read correctly this isn't a PID reserved for EDL mode. Is there some other check we should include here?

If I remember correctly, I did the change as device was not detecting in crash dump mode after crash., there device manager was showing (0x90db as pid) let me recheck this.

andersson:
Yes, that part is correct. But what I'm saying is that 0x90db seems to be used for non-ramdump use cases as well. What I'm looking for is confirmation that we have the right checks in place to not attempt to do ramdump on such alternative cases.

mukeshojha-linux:
I verified again, we don't need this patch, closing it.

PR #223, opened 2026-05-01
For linux msm/program search relative xml

quic-bjorande:


tests/test_program_load_xml.c
```diff
@@ -0,0 +1,404 @@
...
+
+	return remove(path);
+}
+
+static void remove_tree(const char *path)
+{
+	if (path[0] == '\0')
+		return;
+
+	nftw(path, remove_path, 16, FTW_DEPTH | FTW_PHYS);
```
igoropaniuk: is this POSIX only call as well? does it work properly on Windows?

tests/test_program_load_xml.c
```diff
@@ -0,0 +1,404 @@
...
+	op = get_only_op(&ops);
+	ret = snprintf(expected, sizeof(expected), "%s/payload.bin", env->xmldir);
+	assert_true(ret > 0 && (size_t)ret < sizeof(expected));
+	assert_string_equal(op->filename, expected);
+
+	firehose_free_ops(&ops);
+	xmlFreeDoc(doc);
+}
+
+static void test_current_directory_fallback(void **state)
```
igoropaniuk: is a slight misnomer - it doesn't actually test the new path-resolution code, worth either renaming or making the test exercise the new function more directly.

program.c
```diff
@@ -150,12 +150,65 @@ static int program_load_sparse(struct list_head *ops, struct firehose_op *progra
...
+
+	/* Don't attempt to resolve filenames in zip files */
+	if (program->zip)
+		return 0;
+
+	/* Look for the file in include directory */
+	if (incdir) {
+		ret = snprintf(candidate, sizeof(candidate), "%s/%s", incdir, filename);
+		if (ret < 0 || (size_t)ret >= sizeof(candidate))
+			return -ENAMETOOLONG;
```
igoropaniuk: This change currently lives in `program: Extract incdir path resolution`, which is a pure refactor, can we move it to a separate commit?

program.c
```diff
@@ -230,6 +278,15 @@ static int load_program_tag(struct list_head *ops, xmlNode *node, bool is_nand,
 	qdl_file_close(&file);
 
 	return 0;
+
+err_free_op:
```
igoropaniuk, line 279: `qdl_file_close()` as well?

meson.build
```diff
@@ -101,6 +102,31 @@ test(
   protocol: 'tap',
 )
 
+if cmocka_dep.found()
+  test_program_load_xml = executable('test_program_load_xml',
```
igoropaniuk, line 115: we could introduce a dedicated `meson.build` in the `tests/` and move test definitiones there (like how it's done in decouple PR https://github.com/linux-msm/qdl/pull/209)

tests/test_program_load_xml.c
```diff
@@ -0,0 +1,404 @@
...
+static void test_incdir_takes_precedence(void **state)
+{
+	struct test_env *env = *state;
+	struct firehose_op *op;
+	struct list_head ops = LIST_INIT(ops);
+	xmlDoc *doc;
+	char expected[PATH_MAX];
+	int ret;
+
+	make_file(env->incdir, "payload.bin");
```
igoropaniuk: Consider mocking `access()` instead of touching the real filesystem. The whole `program_resolve_path()` function is a thin wrapper around `access()` calls in a specific order, which is exactly the contract a unit test should pin down. Right now the tests verify the result (`op->filename` ends up correct) but can't tell whether `program_resolve_path()` checked incdir then xmldir, or skipped one, or looked at both even when the first hit. cmocka's wrap-based mocking expresses the contract directly:

```
int __wrap_access(const char *path, int mode)
{
    check_expected_ptr(path);
    check_expected(mode);
    return mock_type(int);
}

static void test_incdir_takes_precedence(void **state)
{
    struct test_env *env = *state;
    /* ... build doc ... */

    /* Only the incdir lookup must happen; xmldir must not be probed */
    expect_string(__wrap_access, path, "/incdir/payload.bin");
    expect_value(__wrap_access, mode, F_OK);
    will_return(__wrap_access, 0);  /* hit */

    ret = program_load_xml(&ops, doc, NULL, "/xml/program.xml",
                           false, false, "/incdir");
    assert_int_equal(ret, 0);
    assert_string_equal(get_only_op(&ops)->filename, "/incdir/payload.bin");
}
```

igoropaniuk: This will help us to also drop the POSIX-only filesystem calls (`mkdtemp`, `nftw`, `chdir`)

tests/test_program_load_xml.c
```diff
@@ -0,0 +1,404 @@
...
+		       "num_partition_sectors=\"1\" "
+		       "physical_partition_number=\"0\" "
+		       "start_sector=\"0\" "
+		       "file_sector_offset=\"0\""
+		       "/>"
+		       "</data>",
+		       filename);
+	assert_true(ret > 0 && (size_t)ret < sizeof(xml));
+
+	return xmlReadMemory(xml, ret, "program.xml", NULL, 0);
```
igoropaniuk, line 234: btw, we can introduce `group_setup` / `group_teardown` for `xmlInitParser()` / `xmlCleanupParser()`. 

Currently each test calls xmlReadMemory and xmlFreeDoc, but libxml's global parser state isn't initialized once and torn down once. Minor request - works without it - but it's the cmocka idiom.

tests/test_program_load_xml.c
```diff
@@ -0,0 +1,404 @@
...
+static struct firehose_op *get_only_op(struct list_head *ops)
+{
+	struct firehose_op *op;
+	struct list_head *it;
+	size_t count = 0;
+
+	list_for_each(it, ops)
+		count++;
+
+	assert_int_equal(count, 1);
```
igoropaniuk: nitpick, we can have assertions where we can provide some diagnostic messages `assert_int_equal_msg()`

tests/test_program_load_xml.c
```diff
@@ -0,0 +1,404 @@
...
+}
+
+static void make_file(const char *dir, const char *filename)
+{
+	char path[PATH_MAX];
+	int fd;
+	int ret;
+
+	ret = snprintf(path, sizeof(path), "%s/%s", dir, filename);
+	assert_true(ret > 0 && (size_t)ret < sizeof(path));
```
igoropaniuk: `assert_in_range(ret, 1, sizeof(path) - 1)`

igoropaniuk (commented):
LGTM, just some minor comments/nitpicks

tests/test_program_load_xml.c
quic-bjorande: The test runs on all three OSs, and it was failing on MacOS initially. So, looks like we're good.

tests/test_program_load_xml.c
quic-bjorande: I agree. I'm renaming the test case to clarify the intent, and adding another one to test the precedence rules.

program.c
quic-bjorande: I think this is a good such commit - but the return value check does alter the behavior of the load - and the if (zip) above is a separate commit.

tests/test_program_load_xml.c
quic-bjorande: I'm not able to find any source that tells me that I should call xmlCleanupParser(), other than when using valgrind. Can you point me to some guidelines here?

meson.build
quic-bjorande: Looks like a reasonable continuation of this.

tests/test_program_load_xml.c
igoropaniuk: Found this here https://dev.w3.org/XInclude-Test-Suite/libxml2-2.4.24/libxml2-2.4.24/doc/xmlmem.html. I was expecting that it's public API for libxml2 and is available for the end user (but I might have been mistaken):

> Libxml is not stateless, there is a few set of memory structures needing allocation before the parser is fully functional (some encoding structures for example). This also mean that once parsing is finished there is a tiny amount of memory (a few hundred bytes) which can be recollected if you don't reuse the parser immediately:
    [xmlCleanupParser ()](http://xmlsoft.org/html/libxml-parser.html) is a centralized routine to free the parsing states. Note that it won't deallocate any produced tree if any (use the xmlFreeDoc() and related routines for this).
    [xmlInitParser ()](http://xmlsoft.org/html/libxml-parser.html) is the dual routine allowing to preallocate the parsing state which can be useful for example to avoid initialization reentrancy problems when using libxml in multithreaded applications

andersson: That part I did find, but I don't see how it relates to our test group. Why would we want to invoke these as the group start ends?

I'll merge this and we can add such operations in an incremental change - which will also provide the motivation for doing this; and make it clear why we should replicate this onto each other test set etc.

PR #225, opened 2026-05-04
For linux msm/contents xml flashing

quic-bjorande:
While QDL has been used primarily in use cases where all binaries are collected in a single directory (a so-called "flat build"), Qualcomm builds are typically found with build artifacts scattered across image-specific paths, with a central directory in the form of the "contents.xml" file.

Among other things, this file describes where tools can find programmers, program, and patch files, it describes the mapping to the image-specific paths, etc.

Add support to the newly introduced "qdl flash" to be pointed at a contents.xml file and flash the build.

Then introduce the new "qdl create-zip" subcommand, which instead of flashing the build constructs a self-contained "flashmap zip" file, which can be distributed and passed directly to "qdl flash".

### DIFF
--- a/contents.c
+++ b/contents.c
@@ -1,0 +1,1003 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+#include <ctype.h>
+#include <libgen.h>
+#include <limits.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include <libxml/parser.h>
+#include <libxml/tree.h>
+#include <libxml/xmlstring.h>
+
+#include "contents.h"
+#include "file.h"
+#include "firehose.h"
+#include "pathbuf.h"
+#include "qdl.h"
+
+enum contents_file_type {
+	CONTENTS_FILE_OTHER,
+	CONTENTS_FILE_PROGRAM,
+	CONTENTS_FILE_PATCH,
+	CONTENTS_FILE_DEVICE_PROGRAMMER,
+	CONTENTS_FILE_PROGRAMMER_XML,
+};
+
+enum firehose_type {
+	FIREHOSE_TYPE_NONE,
+	FIREHOSE_TYPE_LITE,
+	FIREHOSE_TYPE_TRUE,
+};
+
+struct contents_entry {
+	enum contents_file_type file_type;
+
+	enum qdl_storage_type storage_type;
+	char *flavor;
+
+	char *filename;
+	struct pathbuf path;
+
+	enum firehose_type firehose_type;
+
+	struct list_head node;
+};
+
+struct contents {
+	struct list_head entries;
+	struct pathbuf base_dir;
+
+	char **flavors;
+	size_t num_flavors;
+};
+
+struct contents_filter {
+	struct contents *contents;
+
+	enum qdl_storage_type storage_type;
+	const char *flavor;
+};
+
+struct contents_selector {
+	enum qdl_storage_type storage_type;
+	const char *flavor;
+};
+
+static const char *contents_storage_name(enum qdl_storage_type storage)
+{
+	const char *name = encode_storage_type(storage);
+
+	return name ? name : "unknown";
+}
+
+static char *contents_node_get_text(xmlNode *node)
+{
+	const xmlChar *start;
+	const xmlChar *end;
+	xmlChar *str;
+	size_t len;
+	char *ret;
+
+	str = xmlNodeGetContent(node);
+	if (!str)
+		return NULL;
+
+	for (start = str; *start && isspace((unsigned char)*start); start++)
+		;
+
+	for (end = start + xmlStrlen(start); end > start &&
+	     isspace((unsigned char)end[-1]); end--)
+		;
+
+	len = end - start;
+	ret = calloc(1, len + 1);
+	if (!ret)
+		goto out_free_str;
+
+	memcpy(ret, start, len);
+
+out_free_str:
+	xmlFree(str);
+
+	return ret;
+}
+
+static bool contents_entry_is_ignored(xmlNode *node)
+{
+	xmlChar *ignore;
+	bool ret;
+
+	ignore = xmlGetProp(node, (xmlChar *)"ignore");
+	if (!ignore)
+		return false;
+
+	ret = !xmlStrcmp(ignore, (xmlChar *)"true");
+
+	xmlFree(ignore);
+
+	return ret;
+}
+
+static int contents_parse_pf(struct contents *contents, xmlNode *node)
+{
+	char **new_flavors;
+	char *name = NULL;
+
+	for (; node; node = node->next) {
+		if (node->type == XML_ELEMENT_NODE &&
+		    !xmlStrcmp(node->name, (xmlChar *)"name")) {
+			name = contents_node_get_text(node);
+			break;
+		}
+	}
+
+	if (!name)
+		return -1;
+
+	new_flavors = realloc(contents->flavors, (contents->num_flavors + 1) * sizeof(*new_flavors));
+	if (!new_flavors) {
+		free(name);
+		return -1;
+	}
+
+	contents->flavors = new_flavors;
+	contents->flavors[contents->num_flavors++] = name;
+
+	return 0;
+}
+
+static int contents_parse_pfs(struct contents *contents, xmlNode *node)
+{
+	int ret;
+
+	for (; node; node = node->next) {
+		if (node->type != XML_ELEMENT_NODE)
+			continue;
+
+		if (!xmlStrcmp(node->name, (xmlChar *)"pf")) {
+			ret = contents_parse_pf(contents, node->children);
+			if (ret < 0) {
+				ux_err("failed to parse product flavor definition\n");
+				return -1;
+			}
+		}
+	}
+
+	return 0;
+}
+
+static enum contents_file_type contents_detect_file_type(xmlNode *node,
+							 enum firehose_type *fh_type)
+{
+	enum contents_file_type type = CONTENTS_FILE_OTHER;
+	xmlChar *firehose_type;
+
+	if (!xmlStrcmp(node->name, (xmlChar *)"partition_file"))
+		type = CONTENTS_FILE_PROGRAM;
+	if (!xmlStrcmp(node->name, (xmlChar *)"partition_patch_file"))
+		type = CONTENTS_FILE_PATCH;
+	if (!xmlStrcmp(node->name, (xmlChar *)"device_programmer"))
+		type = CONTENTS_FILE_DEVICE_PROGRAMMER;
+
+	firehose_type = xmlGetProp(node, (xmlChar *)"firehose_type");
+	if (firehose_type) {
+		if (!xmlStrcmp(firehose_type, (xmlChar *)"true")) {
+			type = CONTENTS_FILE_PROGRAMMER_XML;
+			*fh_type = FIREHOSE_TYPE_TRUE;
+		} else if (!xmlStrcmp(firehose_type, (xmlChar *)"lite")) {
+			*fh_type = FIREHOSE_TYPE_LITE;
+		} else {
+			*fh_type = FIREHOSE_TYPE_NONE;
+		}
+	}
+
+	xmlFree(firehose_type);
+
+	return type;
+}
+
+static enum qdl_storage_type contents_detect_storage_type(xmlNode *node)
+{
+	enum qdl_storage_type type = QDL_STORAGE_UNKNOWN;
+	xmlChar *storage;
+
+	storage = xmlGetProp(node, (xmlChar *)"storage_type");
+	if (storage) {
+		type = decode_storage_type((char *)storage);
+		xmlFree(storage);
+	}
+
+	return type;
+}
+
+static int contents_expand_path_vars(struct pathbuf *path)
+{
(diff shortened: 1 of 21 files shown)


quic-bjorande:
The first 5 patches here are the cleanups and improvements to the relative path already present in #223.
Assuming to be rebasing this ontop of the result of that PR.

program.c
```diff
@@ -230,6 +229,15 @@ static int load_program_tag(struct list_head *ops, xmlNode *node, bool is_nand,
 	qdl_file_close(&file);
 
 	return 0;
+
+err_free_op:
```
igoropaniuk, line 294: `qdl_file_close(&file)` as well?

program.c
```diff
@@ -215,7 +214,7 @@ static int load_program_tag(struct list_head *ops, xmlNode *node, bool is_nand,
 	if (program->filename && program->sparse) {
 		ret = program_load_sparse(ops, program, &file);
```
igoropaniuk, line 276: Seems that partially-appended sparse-chunk ops remain on ops and aren't unwound if error occurs

util.c
```diff
@@ -124,6 +124,44 @@ bool attr_as_bool(xmlNode *node, const char *attr, int *errors)
 	return ret;
 }
 
+const char *encode_storage_type(enum qdl_storage_type storage)
```
igoropaniuk, line 135: can we just have a single table something like:

```
static const struct { enum qdl_storage_type t; const char *name; } storage_types[]
```
driving both functions?

qdl.c
```diff
@@ -452,6 +453,8 @@ static void print_usage(FILE *out)
 	fprintf(out, "          \t\tnumber S, the number of sectors to follow L, or partition by \"name\"\n");
 	fprintf(out, " <ramdump-path>\t\tpath where ramdump should stored\n");
 	fprintf(out, " <segment-filter>\toptional glob-pattern to select which segments to ramdump\n");
+	fprintf(out, " <flashmap>\tflashmap JSON file, or ZIP archive with flashmap.json\n");
+	fprintf(out, " <specifier>\tcomma-separated list of storage specifiers\n");
 	fprintf(out, "\n");
 	fprintf(out, "Example: %s prog_firehose_ddr.elf rawprogram*.xml patch*.xml\n", __progname);
```
igoropaniuk, line 469: could we also add another example besides the legacy way?

```
%s flash contents.xml::ufs,layout0
```

zipper.c
```diff
@@ -0,0 +1,998 @@
...
+		zip_discard(zip);
+		goto out_cleanup;
+	}
+	ux_info("successfully created %s\n", filename);
+
+	ret = 0;
+	goto out_cleanup;
+
+out_discard_zip:
+	zip_discard(zip);
```
igoropaniuk, line 1055: IIUC, on any failure, a partial/corrupt zip is left on disk because `zip_open(filename, ZIP_CREATE | ZIP_TRUNCATE, ...)` truncated it and `zip_discard()` only releases the in-memory archive. 

We can write to `filename.tmp` and `rename(2)` on success; `unlink(tmp)` on every error path.

zipper.c
```diff
@@ -0,0 +1,998 @@
...
+				goto out_discard_zip;
+			}
+
+			file_offset = op->file_offset;
+			tmp_zip_name = NULL;
+			zip_name = NULL;
+
+			if (op->sparse) {
+				switch (op->sparse_chunk_type) {
+				case CHUNK_TYPE_RAW:
```
igoropaniuk, line 940: The commit message claims:

> Sparse files are written to the file by storing the relevant chunks in their decoded form, rather than storing the sparse file (or full unsparsed file). [...] benefits of us not storing DONT_CARE entries, and it avoids the need for seeking as we flash the zip file.

IIUC what the code actually does for `CHUNK_TYPE_RAW`:
`file_offset = (unsigned int)((uint64_t)op->sparse_offset / op->sector_size);`

then falls through to `zipper_register_program_file(... op->filename ...)`, which adds the whole sparse-formatted source file to the zip via `zip_source_file(zip, path, 0, -1)`. Each per-chunk program XML entry references that single shared file with a different file_sector_offset. So DONT_CARE bytes are stored (they're inside the sparse-formatted source).

zipper.c
```diff
@@ -0,0 +1,998 @@
...
+		return 0;
+	}
+
+	file_basename = zipper_basename(source_name);
+	if (!file_basename) {
+		ux_err("failed to parse basename for \"%s\"\n", source_name);
+		return -1;
+	}
+
+	renamed_file = calloc(1, strlen(memory) + strlen(file_basename) + 2);
```
igoropaniuk: nitpick: we can just use asprintf instead of calloc with multiple strlen invocations + snprintf then
```
if (asprintf(&renamed_file, "%s_%s", memory, file_basename) < 0) ..
```

zipper.c
```diff
@@ -0,0 +1,998 @@
...
+	snprintf(storage->patch_xml_name, strlen(storage->memory) + strlen("_patch.xml") + 1,
+		 "%s_patch.xml", storage->memory);
+
+	if (zipper_name_is_used(name_entries, storage->program_xml_name) ||
+	    zipper_name_is_used(name_entries, storage->patch_xml_name)) {
+		ux_err("duplicate xml filename for memory \"%s\"\n", storage->memory);
+		return -1;
+	}
+
+	xmlDocDumpMemory(storage->program_doc, &xml_data, &len);
```
igoropaniuk, line 772: check after `xmlDocDumpMemory`:

```
if (!xml_data || len < 0) {
```

zipper.c
```diff
@@ -0,0 +1,998 @@
...
+	va_start(ap, fmt);
+	va_copy(aq, ap);
+	len = vsnprintf(NULL, 0, fmt, aq);
+	va_end(aq);
+	if (len < 0) {
+		va_end(ap);
+		return;
+	}
+
+	buf = calloc(1, (size_t)len + 1);
```
igoropaniuk: malloc() is sufficient, as vsnprintf overwrites the buffer after this calls on line 63

zipper.c
```diff
@@ -0,0 +1,998 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ */
+#include <limits.h>
+#include <libgen.h>
+#include <stdarg.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
```
igoropaniuk, line 14: Can we add LFS preamble that firehose.c, program.c, sparse.c, gpt.c have already:

```
#define _FILE_OFFSET_BITS 64
```

The file consumes `op->sparse_offset` which is off_t and calls zip_source_file(zip, path, 0, -1) on potentially-multi-GB images

meson.build
```diff
@@ -141,6 +142,41 @@ if cmocka_dep.found()
     protocol: 'tap',
     env: ['CMOCKA_MESSAGE_OUTPUT=TAP'],
   )
+
+  test_contents_selectors = executable('test_contents_selectors',
```
igoropaniuk, line 155: As I mentioned before, let's move it to a dedicated meson.build in `tests/ subdirectory

contents.c
```diff
@@ -0,0 +1,962 @@
...
+			ret = patch_load(ops, qdl_pathbuf_str(&entry->path));
+			if (ret < 0) {
+				ux_err("failed to load %s\n", entry->filename);
+				goto out_free_contents;
+			}
+		}
+	}
+
+out_free_contents:
+	list_for_each_entry_safe(entry, next, &contents.entries, node) {
```
igoropaniuk, line 938: cleanup here might crash as `list_init()` might not be called, for example in contents_load_xml():

```
int contents_load_xml(struct contents *contents, const char *filename)
{
	xmlNode *root;
	xmlDoc *doc;
	int ret;

	doc = xmlReadFile(filename, NULL, 0);
	if (!doc) {
		ux_err("failed to parse contents file \"%s\"\n", filename);
		return -1;
	}

	root = xmlDocGetRootElement(doc);
	if (!root || xmlStrcmp(root->name, (xmlChar *)"contents")) {
		ux_err("specified file \"%s\" is not a contents.xml document\n", filename);
		xmlFreeDoc(doc);
		return -1;
	}

	ret = contents_get_base_dir(&contents->base_dir, filename);
	if (ret < 0)
		goto err_free_doc;

	list_init(&contents->entries);       <------- is called only here, but there are 3 failure points above
```

contents.c
```diff
@@ -0,0 +1,962 @@
...
+static int contents_parse_pf(struct contents *contents, xmlNode *node)
+{
+	const char **new_flavors;
+	const char *name;
+
+	for (; node; node = node->next) {
+		if (node->type != XML_ELEMENT_NODE)
+			continue;
+
+		name = contents_node_get_text(node);
```
igoropaniuk: can return NULL

contents.c
```diff
@@ -0,0 +1,962 @@
...
+
+		new_flavors = realloc(contents->flavors, (contents->num_flavors + 1) * sizeof(*new_flavors));
+		if (!new_flavors)
+			return -1;
```
igoropaniuk: leaks `name`

contents.c
```diff
@@ -0,0 +1,962 @@
...
+			}
+		}
+	}
+
+out_free_contents:
+	list_for_each_entry_safe(entry, next, &contents.entries, node) {
+		free(entry->filename);
+		free(entry->flavor);
+		free(entry);
+	}
```
igoropaniuk, line 942: should we also free contents.flavors?

```
for (i = 0; i < contents.num_flavors; i++)
    free((void *)contents.flavors[i]);
free(contents.flavors);
```

contents.c
```diff
@@ -0,0 +1,962 @@
...
+
+	for (selector_idx = 0; selector_idx < count; selector_idx++) {
+		if (selectors[selector_idx].storage_type == storage_type)
+			return true;
+	}
+
+	return false;
+}
+
+static int contents_collect_valid_selectors(struct contents *contents,
```
igoropaniuk: static size_t, as below it returns size_t?

contents.c
```diff
@@ -0,0 +1,962 @@
...
+		storage_type = selectors[i].storage_type;
+		flavor = selectors[i].flavor;
+
+		op = firehose_alloc_op(FIREHOSE_OP_CONFIGURE);
+		op->storage_type = storage_type;
+		list_append(ops, &op->node);
+
+		list_for_each_entry(entry, &contents.entries, node) {
+			if (entry->file_type != CONTENTS_FILE_PROGRAM)
+				continue;
```
igoropaniuk, line 899: can we add some debug output here instead of swallowing these failures with `continue`?

contents.c
```diff
@@ -0,0 +1,962 @@
...
+		}
+
+		storage = decode_storage_type(token);
+		if (storage == QDL_STORAGE_UNKNOWN) {
+			ux_err("unknown storage type \"%s\"\n", token);
+			goto err_free_selectors;
+
+		}
+
+			if (!contents_selector_is_known(valid_selectors, num_valid_selectors, storage, flavor)) {
```
igoropaniuk: indentation issue

contents.c
```diff
@@ -0,0 +1,962 @@
...
+		name = contents_node_get_text(node);
+
+		new_flavors = realloc(contents->flavors, (contents->num_flavors + 1) * sizeof(*new_flavors));
+		if (!new_flavors)
+			return -1;
+
+		contents->flavors = new_flavors;
+		contents->flavors[contents->num_flavors++] = name;
+
+		return 0;
```
igoropaniuk: always returns from inside loop? why?

contents.c
```diff
@@ -0,0 +1,962 @@
...
+#include <string.h>
+#include <unistd.h>
+
+#include <libxml/parser.h>
+#include <libxml/tree.h>
+
+#include "contents.h"
+#include "file.h"
+#include "firehose.h"
+#include "libxml/xmlstring.h"
```
igoropaniuk: please sort includes

igoropaniuk (commented):
added some comments

contents.c
quic-bjorande: This isn't a failure, the loop is scanning all entries and skipping items that aren't CONTENTS_FILE_PROGRAM, doesn't have valid storage type or doesn't have storage and flavor matching the selector we're currently scannning for.

Simply inverse of the thing we're actually looking for, to save us 3 levels of indentation.

contents.c
quic-bjorande: The intent was that we're just trying to fish out the "text" of the first child tag of the <pf>. But upon closer look, we should really make sure to pull the "text" of specifically the <name> tag. So, I rewrote this to look for one <name>.

contents.c
quic-bjorande: Also extended this to deal with the case where we find no viable selectors (and treaded both as an error)

zipper.c
quic-bjorande: We can actually asprintf() this one even.

zipper.c
quic-bjorande: Thank you!

zipper.c
quic-bjorande: I implemented the suggestion. Sometimes I see libzip leave .part files around when I interrupt it. Let's tackle that in a follow up series.

program.c
quic-bjorande: This comment applies to the already merged change, we should clean this up separately.

program.c
quic-bjorande: Fixed this in the other PR

PR #226, opened 2026-05-05
patch: suppress libxml2 stderr output on parse errors

igoropaniuk:
libxml2 prints warnings and errors directly to stderr before the caller gets a chance to handle them. This leaks raw parser diagnostics into both test output and user-facing output when a patch file is missing or malformed. patch_load() already checks the return value and reports the failure through ux_err(), so the extra stderr output is redundant noise.

Pass XML_PARSE_NOWARNING | XML_PARSE_NOERROR to xmlReadFile() so all error reporting goes through our own output layer.

### DIFF
--- a/patch.c
+++ b/patch.c
@@ -70,7 +70,7 @@ int patch_load(struct list_head *ops, const char *patch_file)
 	xmlDoc *doc;
 	int ret;
 
-	doc = xmlReadFile(patch_file, NULL, 0);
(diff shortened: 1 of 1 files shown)


igoropaniuk:
@andersson the fix for patch we've discussed recently
I've been thinking using the same flags everywhere in all  xmlReadFile() invocations (let me know what you think)

PR #230, opened 2026-05-06
tests: stop fixture generation from timing out CI on slow runners

igoropaniuk:
The integration tests' fixture generator was writing ~1 GB of zeros to disk
(efi.bin = 512 MB, rootfs.img = 500 MB) and zipping it on every test run.
That pushed both `qdl:Flashmap dry-run` and `qdl:VIP table generation` past
meson's default 30 s timeout on the slow `macos-15-intel` GitHub runner.

Two commits address it:

* Switch `dd` to `count=0 seek=N` so the backing files are sparse - reads
  still yield zeros, but allocation and write I/O are O(1).
* Shrink `efi.bin` and `rootfs.img` to 8 MB each (the tests don't care about
  the contents, only that qdl parses the descriptors and walks the program
  list). `rawprogram0.xml` and `EXPECTED_DIGEST` are updated to match.

Combined local test time drops from ~7 s to ~0.25 s; the slow Intel runner
should now finish well under the 30 s timeout.


igoropaniuk:
@andersson I have also two more options how we can address that:

1) Increase timeout to 120s for meson tests
2) Deprecate MacOs Intel runner

tests/data/generate_flat_build.sh
```diff
@@ -10,7 +10,12 @@ create_file_with_size() {
 	filename="$1"
 	size_kbytes="$2"
 
-	dd if=/dev/zero of="$OUTDIR/$filename" bs=1024 count="$size_kbytes" status=none
+	# Create a sparse file of the requested size: dd with count=0 seek=N
+	# extends the file to N*bs bytes via lseek without writing any data.
+	# Reads still yield zeros, so callers (sha256, qdl --dry-run, zip) see
+	# the same content as a fully-written file, but disk I/O and allocation
+	# are O(1). Avoids ~1 GB of writes on slow CI runners.
+	dd if=/dev/zero of="$OUTDIR/$filename" bs=1024 count=0 seek="$size_kbytes" status=none
```
quic-bjorande, line 18: Can't we use "truncate -s $size_kbytes $OUTDIR/$filename" instead?

With a comment saying something like "Inflate the file to $size_kbytes. Content will be read as all zeros by sha256, qdl, zip etc."?

PR #232, opened 2026-05-07
Add --skip-reset and auto-skip Sahara when programmer is already running

igoropaniuk:
Support workflows where the Firehose programmer is left running between qdl invocations - useful when chaining flash operations, doing post-flash verification, or running a series of reads/writes against the same programmer session.

`--skip-reset (-R)` suppresses the unconditional reset issued at the end of `firehose_run()`, leaving the programmer responsive after the command completes.

A second invocation against the still-running programmer no longer needs an explicit flag: sahara_run() now peeks at the device's first response and, if it sees a Firehose XML banner (`<?xml`) or a read timeout instead of a Sahara HELLO, returns success so the caller proceeds straight to the Firehose phase. This mirrors qdlrs's behaviour and avoids the need for a separate `--skip-sahara` flag.

Typical use:

```bash
$ qdl prog_firehose_ddr.elf rawprogram*.xml patch*.xml --skip-reset
$ qdl prog_firehose_ddr.elf write 0 fastboot.img    # Sahara auto-skipped
```

### DIFF
--- a/firehose.c
+++ b/firehose.c
@@ -1053,7 +1053,7 @@ static int firehose_detect_and_configure(struct qdl_device *qdl,
 	return 0;
 }
 
-int firehose_provision(struct qdl_device *qdl)
+int firehose_provision(struct qdl_device *qdl, bool skip_reset)
 {
 	int ret;
 
@@ -1069,7 +1069,8 @@ int firehose_provision(struct qdl_device *qdl)
 	else
 		ux_info("UFS provisioning failed\n");
 
-	firehose_reset(qdl);
+	if (!skip_reset)
+		firehose_reset(qdl);
 
 	return ret;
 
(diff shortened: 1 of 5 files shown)


loicpoulain:
This looks promising. However, I think qdl should remain predictable, with consistent step validation, as it is widely used in both testing and production environments. The firehose autodetection, relying on USB timeouts or XML payload may be somewhat fragile. *In my opinion*, it would be preferable to introduce a simple --skip-sahara flag for now rather than relying on this autodetection mechanism.

igoropaniuk:
@loicpoulain the motivation why automatic detection was added https://github.com/linux-msm/qdl/pull/201 (@andersson comment)

quic-bjorande:
> introduce a simple --skip-sahara flag

The problem I have with this is that you're now putting the burden on the user to understand the underlying protocols; it's no longer a flashing tool, it's a Sahara+something-else-tool.

> qdl should remain predictable

This is the reason why I haven't introduced this feature in the past, although I've implemented it several times. I think as proposed it fits nicely into the case of a series of "qdl read/write" spanning multiple operations, with little risk of unpredictability. But, if you hit ^C in the middle of a <program> or <read>, you will not have a predictable continuation...

So, I think this is good usability, but I certainly share your concern and am hesitant to hit the merge button...

andersson:
How about us making the reset a firehose_op that we append in qdl.c. Then make that append conditional on the basis of this flag?

igoropaniuk:
> How about us making the reset a firehose_op that we append in qdl.c. Then make that append conditional on the basis of this flag?

@andersson 
done, thanks for the suggestion!

PR #233, opened 2026-05-11
qdl: add QUD backend so on Windows flashing via the official QDLoader 9008 driver works

igoropaniuk:
On Windows, the official Qualcomm Userspace Driver / QDLoader 9008 kernel-mode driver claims the EDL device's USB interface during PnP enumeration and exposes it as `\\.\COMx`. `libusb` can't reach the bulk endpoints in that state, so today the only way to flash with qdl on Windows is to swap the binding to `WinUSB` (for example via Zadig - which is awkward, sometimes blocked by IT policy, and breaks any other QC tooling that expects the COM-port view.

This PR adds a second transport ("QUD" - Qualcomm USB Driver) that talks to the device through the kernel-mode driver's COM-port file directly, leaving Sahara/Firehose framing untouched. Linux is untouched and continues to use libusb.

### DIFF
--- a/auto.c
+++ b/auto.c
@@ -1,0 +1,165 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
+ *
+ * QDL_DEVICE_AUTO: meta-backend that defers transport selection to the
+ * wait loop. Each 250 ms tick it runs both the libusb open attempt and
+ * (on Windows) a QUD SetupAPI probe; whichever first reaches an EDL
+ * device wins, and its concrete qdl_device is bound as the inner. All
+ * subsequent qdl_read/write/close calls on the outer forward to the
+ * inner.
+ */
+
+#include <errno.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "qdl.h"
+
+struct qdl_device_auto {
+	struct qdl_device base;
+	struct qdl_device *inner;
+	long pending_chunk_size;
+	bool chunk_size_set;
+};
+
+static struct qdl_device_auto *to_auto(struct qdl_device *qdl)
+{
+	return container_of(qdl, struct qdl_device_auto, base);
+}
+
+static void auto_bind_inner(struct qdl_device_auto *wrap, struct qdl_device *inner)
+{
+	wrap->inner = inner;
+	wrap->base.max_payload_size = inner->max_payload_size;
+	if (wrap->chunk_size_set)
+		inner->set_out_chunk_size(inner, wrap->pending_chunk_size);
+}
+
+static int auto_open(struct qdl_device *qdl, const char *serial)
+{
+	struct qdl_device_auto *wrap = to_auto(qdl);
+	struct qdl_device *usb_dev;
+#ifdef _WIN32
+	struct qdl_device *qud_dev;
+	int qud_count;
+#endif
+	int visible_prev = -1;
+	int visible;
+	int ret;
+
+	usb_dev = usb_init();
+	if (!usb_dev)
+		return -1;
+
+#ifdef _WIN32
+	qud_dev = qud_init();
+	if (!qud_dev) {
+		qdl_deinit(usb_dev);
+		return -1;
+	}
+#endif
+
+	for (;;) {
+		ret = try_usb_open(usb_dev, serial, &visible);
+		if (ret == 0) {
+#ifdef _WIN32
+			qdl_deinit(qud_dev);
+#endif
+			auto_bind_inner(wrap, usb_dev);
+			return 0;
+		}
+		if (ret == -EIO)
+			goto fail;
+
+#ifdef _WIN32
+		qud_count = qud_probe_present();
+		if (qud_count > 0 || ret == -EBUSY) {
+			if (qud_dev->open(qud_dev, serial) == 0) {
+				qdl_deinit(usb_dev);
+				auto_bind_inner(wrap, qud_dev);
+				return 0;
+			}
+		}
+		visible += qud_count;
+#endif
+
+		if (visible != visible_prev) {
+			if (visible == 0)
+				ux_info("Waiting for EDL device\n");
+			else if (serial)
+				ux_info("%d EDL device(s) visible, none match serial \"%s\"\n",
+					visible, serial);
+			else
+				ux_info("%d EDL device(s) visible, none could be opened\n",
+					visible);
+			visible_prev = visible;
+		}
+
+		usleep(250000);
+	}
+
+fail:
+	qdl_deinit(usb_dev);
+#ifdef _WIN32
+	qdl_deinit(qud_dev);
+#endif
+	return -1;
+}
+
+static int auto_read(struct qdl_device *qdl, void *buf, size_t len, unsigned int timeout)
+{
+	struct qdl_device *inner = to_auto(qdl)->inner;
+
+	return inner->read(inner, buf, len, timeout);
+}
+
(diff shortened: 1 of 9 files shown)


firehose.c
```diff
@@ -179,15 +179,46 @@ static int firehose_read(struct qdl_device *qdl, int timeout_ms,
...
+
+		while (cursor < bufend) {
+			char *start = strstr(cursor, "<?xml");
+			char *next;
+			size_t chunk;
+
+			if (!start)
+				break;
+
+			next = strstr(start + 5, "<?xml");
```
andersson: When we issue a <read> we'll get a rawmode=true response, which might be immediately followed by the binary data - now in the same read. So I think you need to look for </data> instead.

andersson:
I don't think it's good UX to rely on the user to tell us which USB driver they have installed. In my work-in-progress, I integrated the QUD support directly into usb.c to make it try both on Windows - to make this choice fully transparent... Code is a little bit messier still, which is why I haven't posted it, but UX is much better.

Remember, we don't want the user to have to know how flashing works to flash the device.

igoropaniuk:
> Remember, we don't want the user to have to know how flashing works to flash the device.

By default, the "backend" is selected automatically ("auto"). The--backend CLI option is needed to override this selection and choose manually.

firehose.c
igoropaniuk: Nice catch, we likely will hit this when dumping data from a device (I tested only flashing)

quic-bjorande:
> By default, the "backend" is selected automatically ("auto"). The--backend CLI option is needed to override this selection and choose manually.

Right, but unless I'm reading the code incorrectly, decode_backend("auto") will return QDL_DEVICE_QUD on Windows, regardless of me having the QUD installed or not.

So, a Windows user that can run QDL today, will face exactly the same problem that the QUD-users are facing today, and they will have to know that they need to say "--backend usb" to get things working again.

Similarly, but hopefully a less frequent problem, is that you will have QUD-users who connect the USB cable and then go "--backend usb", because that sounds like the obvious choice.

firehose.c
quic-bjorande: "qdl firehose.elf read 0/1+32 gpt.bin" should give you a GPT header...

igoropaniuk: fixed, thanks!

igoropaniuk:
@andersson added proper auto resultion of the backend:

1. Probes libusb. If a Qualcomm EDL device is visible AND
    libusb_open() succeeds on it, pick QDL_DEVICE_USB. This keeps
    every WinUSB user on the existing libusb path.
2. Otherwise probes the QUD SetupAPI enumeration. If a Qualcomm
     QDLoader COM port is present, OR libusb saw the device but
     could not open it (the no-driver / QUD-claimed signal),
     pick QDL_DEVICE_QUD.
3. Otherwise polls both probes for up to ~2 seconds, to catch the
     common case of the user plugging the cable in just after launch.
     If the bus is still empty, fall back to QDL_DEVICE_USB so the
     existing "Waiting for EDL device" loop keeps running.

andersson:
This is better, but I think the auto-probe timeout is going to come back and bit us.
The technical solution certainly has its merits, but it is not good UX and the average user won't agree.

I think we need to make the USB probe loop check both cases - even though it means that this can't be a clean "backend".

igoropaniuk:
@andersson I've completely reworked the usb backend selection based on your suggestions and also rebased on the latest master. Could you take a look?

auto.c
```diff
@@ -0,0 +1,170 @@
...
+	wrap->base.max_payload_size = inner->max_payload_size;
+	if (wrap->chunk_size_set)
+		inner->set_out_chunk_size(inner, wrap->pending_chunk_size);
+}
+
+static int auto_open(struct qdl_device *qdl, const char *serial)
+{
+	struct qdl_device_auto *wrap = to_auto(qdl);
+	struct qdl_device *usb_dev;
+#ifdef _WIN32
```
igoropaniuk, line 45: we can drop this _WIN32 later, when there is support for Linux QUD driver

qud.c
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright (c) 2026, Qualcomm Innovation Center, Inc. All rights reserved.
```
quic-bjorande: Please use the new copyright statement.

io.c
```diff
@@ -48,14 +52,80 @@ void qdl_close(struct qdl_device *qdl)
...
+		qdl->pending_len = len;
+		qdl->pending_off = 0;
+		return 0;
+	}
+
+	leftover = qdl->pending_len - qdl->pending_off;
+	grown = malloc(leftover + len);
+	if (!grown)
+		return -ENOMEM;
+	memcpy(grown, qdl->pending_buf + qdl->pending_off, leftover);
```
quic-bjorande: I think this would be cleaner with a realloc and then use memmove to shift the data from pending_off to 0.

auto.c
```diff
@@ -0,0 +1,170 @@
+// SPDX-License-Identifier: BSD-3-Clause
+/*
+ * Copyright (c) 2026, Qualcomm Innovation Center, Inc. All rights reserved.
```
quic-bjorande: Copyright statement.

auto.c
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * QDL_DEVICE_AUTO: meta-backend that defers transport selection to the
+ * wait loop. Each 250 ms tick it runs both the libusb open attempt and
+ * (on Windows) a QUD SetupAPI probe; whichever first reaches an EDL
+ * device wins, and its concrete qdl_device is bound as the inner. All
+ * subsequent qdl_read/write/close calls on the outer forward to the
+ * inner.
+ *
+ * This replaces an earlier "decide upfront with a 2 s grace window"
```
quic-bjorande: This comment refers to a design we only had in the PR, but once we commit this PR I think the comment will be confusing.

qdl.h
```diff
@@ -113,6 +121,22 @@ int qdl_vip_transfer_enable(struct qdl_device *qdl, const char *vip_table_path);
...
+ * Returns 0 on success (and emits the "Flashing/Collecting device" UX
+ * line), -ENODEV when no EDL device is visible, -EBUSY when one is
+ * visible but cannot be opened (typically: the Qualcomm QDLoader 9008
+ * driver has claimed it), -EIO on a libusb failure. @visible_out, if
+ * non-NULL, receives the count of EDL devices seen on this pass.
+ *
+ * qud_probe_present() returns the number of Qualcomm COM ports the QUD
+ * backend enumerated via SetupAPI; 0 on non-Windows hosts.
+ */
+int usb_open_attempt(struct qdl_device *qdl, const char *serial, int *visible_out);
```
quic-bjorande: I think it would be more idiomatic to use the name try_usb_open();

auto.c
```diff
@@ -0,0 +1,170 @@
...
+			qdl_deinit(qud_dev);
+#endif
+			auto_bind_inner(wrap, usb_dev);
+			return 0;
+		}
+		if (ret == -EIO)
+			goto fail;
+
+#ifdef _WIN32
+		qud_count = qud_probe_present();
```
quic-bjorande, line 78: Been jumping back and forth in the patch here... It would be nice if the try_usb_open() and the qud path were symmetrical, but I haven't convinced myself if that's possible or not.

quic-bjorande (commented):
This looks much better and looks to provide consistent UX independently of which driver you have installed.

Can we get "qdl list" working as well? To avoid the immediate feature request/bug report.

qud.c
igoropaniuk: fixed

igoropaniuk:
@andersson addressed all your comments except "Been jumping back and forth in the patch here... It would be nice if the try_usb_open() and the qud path were symmetrical, but I haven't convinced myself if that's possible or not."

PR #238, opened 2026-05-12
Add support for cpio and multi-elf in zip archives.

nrclark-gm:
Teach flashmaps how to contain cpio and multi-elf programmers so that it has parity with the command-line options available to qdl.

In addition to the existing legal values:
```json
{"programmer": ["prog_firehose_ddr.elf"]}
```

This allows for the following items to be valid values for "programmer":

```json
{"programmer": ["combo.cpio"]}
```
```json
{"programmer": ["13:foo.elf", "14:bar.elf"]}
```

which makes it possible to use flash-bundles when targeting chips like Nord that need to load multi-stage Firehose programmers.

### DIFF
--- a/file.c
+++ b/file.c
@@ -183,6 +183,7 @@ int qdl_zip_open(const char *filename, struct qdl_zip **__qdl_zip)
 	qdl_zip = calloc(1, sizeof(*qdl_zip));
 	if (!qdl_zip) {
 		zip_close(zip);
+		*__qdl_zip = NULL;
 		return -1;
 	}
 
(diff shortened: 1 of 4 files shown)


andersson:
"qdl create-zip" will create flashmap zip files with multiple programmers defined (if the source material define it as such).

I think rather than supporting the multiple mechanisms for storing multiple programmers in the flashmap.json, we should resolve the xml-format and expand the cpio format into the one supported. Let me know if you see any concerns with this approach.

Issue #237, opened 2026-05-11
Feature request: add support for sha256 retrieval


nrclark-gm:
We do a lot of flashing in our automated-test pipelines, and sometimes it takes as long as the actual test suite being executed. We could speed it up a lot by making use of Firehose's `<getsha256digest>` command. 

I could see us making good use of two different ways to utilize it:

### 1. `qdl sha256` subcommand ###
 
This would be a `sha256` subcommand with the same syntax as `qdl read`, except that it sends `<getsha256digest>` instead of `<read>` and prints the ascii-formatted sha256 to stdout.

We'd tend to use this at the bench to answer questions like "what's flashed onto my board?", which is something that can be really helpful on shared boards or lab machines.

and/or:

### 2. `qdl --skipblock` argument ###

This would add a `--skipblock=sha256` command-line argument to `qdl` that would enable the following behavior:

Before each `<program>` operation, a `<getsha256digest>` would be sent to fetch the checksum of the target portion of flash memory. If the returned checksum matches the locally-calculated value, then the `<program>` operation(s) would get skipped.

This behavior isn't bulletproof since hash collisions are always a possibility, so the default mode would stay as `--skipblock=none`. But `--skipblock=sha256` would be accurate enough for developer use, and could significantly improve total flashing speed (especially on slow storage like `spinor`).

quic-bjorande:
First part of this request has been merged.

igoropaniuk:
@nrclark-gm the feature has been implemented and merged (https://github.com/linux-msm/qdl/pull/242), will be happy to hear your feedback

igoropaniuk:
Closing this issue as it was implemented in 2.7, with additional improvements also included in the latest `master`.

Please feel free to reopen it if you have any additional comments or feedback.

PR #239, opened 2026-05-13
[RFC] qdl: add sha256 subcommand

igoropaniuk:
Expose the FIREHOSE_OP_GET_SHA256_DIGEST primitive as a user-facing
verb. The syntax mirrors the existing read / write / erase verbs and
reuses parse_storage_address(), so addresses can be expressed as
`<P>/<S>+<L>`, by GPT partition name, or any of the other supported
forms.

Each invocation prints one lowercase 64-character hex digest per region
to stdout, so the output can be diffed against a sha256sum(1) line or
piped into other tooling without further parsing.

Typical use is verifying that what is on the device matches a known
local image - useful on shared bench machines where multiple people
flash the same board:

```bash
$ qdl prog_firehose_ddr.elf sha256 xbl_a
...
c288e50a94687a130a64db608c83678777ce5e7f9229248696c6f0dc41406cc9

$ qdl prog_firehose_ddr.elf read xbl_a xbl_a.bin
...
$ sha256sum xbl_a.bin
c288e50a94687a130a64db608c83678777ce5e7f9229248696c6f0dc41406cc9  xbl_a.bin
```

Multiple sha256 verbs can be chained in a single invocation, with one
digest line emitted per region in order.

Link: https://github.com/linux-msm/qdl/issues/237

### DIFF
--- a/firehose.c
+++ b/firehose.c
@@ -25,6 +25,7 @@
 #include "qdl.h"
 #include "file.h"
 #include "firehose.h"
+#include "sha2.h"
 #include "ufs.h"
 #include "oscompat.h"
 #include "vip.h"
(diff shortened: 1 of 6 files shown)


igoropaniuk:
This PR implements the first suggestion from https://github.com/linux-msm/qdl/issues/237.

I plan to create a separate PR for  `--skipblock` support

firehose.c
```diff
@@ -784,6 +866,67 @@ static int firehose_read_op(struct qdl_device *qdl, struct firehose_op *op)
...
+
+	op->digest_valid = false;
+
+	ret = firehose_write(qdl, doc);
+	if (ret < 0) {
+		ux_err("failed to send getsha256digest command\n");
+		goto out;
+	}
+
+	ret = firehose_read(qdl, 30000, firehose_sha256_parser, op);
```
quic-bjorande: Passing and modifying the op itself to get the result back from the parser doesn't look so nice; "op" is effectively the commands list.

I'd rather see a dedicated "result struct" for this.

I guess the open question is how we tie this into the next step?

quic-bjorande: One case where I can think of it being valid to store the data back "into" the op is if the op carries a pointer to a buffer in the calling tool, so upon running firehose_run() that buffer is filled in and upon return one can read it from there - e.g. for printing it out.

But I don't know if we want to go that far, the proposed solution solves the problem in a minimal fashion (but then I don't think we should modify the op).

igoropaniuk: > I'd rather see a dedicated "result struct" for this.

done!

igoropaniuk: I have WIP implementation of the whole feature here https://github.com/igoropaniuk/qdl/tree/feat/skipblock_support_wip, but I haven't yet tested it on different use cases (just flashing meta-qcom build)

The implementation is obviously estricted to non-sparse, non-NAND programs with VIP disabled. Sparse chunks would each need their own digest and NAND has spare/OOB bytes whose semantics differ across.

andersson:
Looking at the continuation of this PR, I think it might be cleaner after all to move back to carrying the digest in the op, and move the print to qdl.c after returning from firehose_run() - as firehose.c deals with protocol stuff and qdl (more and more) defines the policy.

But we can change that around, so I merged this in the meantime.
Thank you.

PR #242, opened 2026-05-15
qdl: add --skipblock=sha256 to skip programs already on flash

igoropaniuk:
Automated flashing pipelines often re-program identical content on every
run, which dominates total flash time on slow storage like spinor. With
the <getsha256digest> primitive now exposed, qdl can ask the device for
the SHA-256 of the target region before each <program> and skip the
write when it already matches what would have been sent.

Add `--skipblock=<none|sha256>`, defaulting to none. When sha256 is
selected, firehose_program() hashes the bytes it would otherwise stream
to the device - reading `num_partition_sectors * SECTOR_SIZE_IN_BYTES`
from the file with zero-padding for the trailing partial sector, exactly
matching what the existing chunked write path puts on the wire - then
asks the device for the digest of the same region via
`firehose_getsha256digest()`, reusing the op->digest field that the
sha256 subcommand path also reads. On a match, the `<program>` is skipped
entirely and "skipped \"<label>\" (sha256 match)" is logged in place of
the usual flashed-at-rate line.

The optimization is gated on three conditions, all enforced silently:

  - !program->sparse: each sparse chunk would need its own digest;
    keeping sparse images on the normal program path avoids per-chunk
    round trips and the bookkeeping that would come with them.
  - !program->is_nand: NAND spare/OOB byte semantics differ across
    programmers, so the digest the device returns is not guaranteed to
    correspond to the bytes qdl would compute locally.
  - VIP disabled: `<getsha256digest>` is not part of pre-built VIP
    digest tables, so issuing it under VIP would cause a hash mismatch
    on the device.

Any failure on the optimization path - file read error, query failure,
digest absent, or digest mismatch - falls through to the normal program
flow. The file position is restored unconditionally by the existing
`qdl_file_seek()` at the top of the write loop, so the fall-through path
is byte-identical to `--skipblock=none`.

Hash collisions are theoretically possible, so the default stays at
none. sha256 is offered for developer-facing iteration where collision
risk is acceptable and the latency savings are large.

Link: https://github.com/linux-msm/qdl/issues/237

### DIFF
--- a/firehose.c
+++ b/firehose.c
@@ -595,6 +595,88 @@ out:
 	return ret == FIREHOSE_ACK ? 0 : -1;
 }
 
+static int firehose_getsha256digest(struct qdl_device *qdl, struct firehose_op *op);
+
+/*
+ * Decide whether the @program op can be skipped because its bytes are
+ * already on flash. Returns 1 if the program path can be bypassed, 0
+ * to defer to the normal write. @buf is a scratch buffer of size
+ * qdl->max_payload_size owned by the caller.
+ *
+ * Hash the bytes that would have been written, ask the device for the
+ * digest of the same region, and compare. Any failure on the way (read error,
+ * digest query failure, mismatch) returns 0 so the caller proceeds with
+ * the normal program flow; the file pointer is restored by the
+ * qdl_file_seek() at the top of the write loop.
+ *
+ * Restricted to non-sparse, non-NAND programs with VIP disabled:
+ *   - Sparse chunks would each need their own digest; v1 keeps them on
+ *     the normal program path.
+ *   - NAND has spare/OOB bytes whose semantics differ across
+ *     programmers.
+ *   - <getsha256digest> is not part of pre-built VIP digest tables.
+ */
+static int qdl_should_skipblock(struct qdl_device *qdl,
+				struct firehose_op *program,
+				struct qdl_file *file,
+				unsigned int num_sectors,
+				unsigned int sector_size,
+				void *buf)
+{
+	uint8_t local_digest[SHA256_DIGEST_LENGTH];
+	size_t total = (size_t)num_sectors * sector_size;
+	struct firehose_op digest_op;
+	size_t hashed = 0;
+	SHA2_CTX ctx;
+	ssize_t hn;
+
+	if (qdl->skipblock_mode != QDL_SKIPBLOCK_SHA256 ||
+	    program->sparse ||
+	    program->is_nand ||
+	    qdl->vip_data.state != VIP_DISABLED)
+		return 0;
+
+	SHA256Init(&ctx);
+	qdl_file_seek(file, (off_t)program->file_offset * sector_size, SEEK_SET);
+
+	while (hashed < total) {
+		size_t want = MIN(qdl->max_payload_size, total - hashed);
+
+		hn = qdl_file_read_exact(file, buf, want);
+		if (hn < 0)
+			return 0;
+		if (hn == 0 || (size_t)hn < want) {
+			/*
+			 * EOF before completing the requested region: the
+			 * device's digest is over `total` bytes from flash,
+			 * so we can't construct a matching local digest.
+			 * Defer to the normal program path.
+			 */
+			if (hn > 0)
+				SHA256Update(&ctx, buf, hn);
+			return 0;
+		}
+		SHA256Update(&ctx, buf, hn);
+		hashed += (size_t)hn;
+	}
+
+	SHA256Final(local_digest, &ctx);
+
+	digest_op = (struct firehose_op){
+		.type = FIREHOSE_OP_GET_SHA256_DIGEST,
+		.sector_size = sector_size,
+		.num_sectors = num_sectors,
+		.partition = program->partition,
+		.start_sector = program->start_sector,
+	};
+
+	if (firehose_getsha256digest(qdl, &digest_op) == 0 &&
+	    !memcmp(local_digest, digest_op.digest, SHA256_DIGEST_LENGTH))
+		return 1;
+
+	return 0;
+}
+
 static int firehose_program(struct qdl_device *qdl, struct firehose_op *program)
 {
 	unsigned int num_sectors;
@@ -649,6 +731,13 @@ static int firehose_program(struct qdl_device *qdl, struct firehose_op *program)
 	if (!buf) {
 		ux_err("failed to allocate sector buffer\n");
 		goto err_close_fd;
+	}
+
+	if (qdl_should_skipblock(qdl, program, &file, num_sectors, sector_size, buf)) {
+		ux_info("skipped \"%s\" (sha256 match)\n", program->label);
+		free(buf);
+		qdl_file_close(&file);
+		return 0;
 	}
 
 	doc = xmlNewDoc((xmlChar *)"1.0");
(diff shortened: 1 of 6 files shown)


igoropaniuk:
@andersson Reworked based on your comment from https://github.com/linux-msm/qdl/pull/239#issuecomment-4456424018

mwasilew:
I tested it in my staging LAVA instance with Glymur. Flashing spinor with identical firmware takes 9 seconds instead of nearly 6 minutes!
Before:
```
start: 2.5 flash-qdl (timeout 00:14:35) [common]
Calling: ['/usr/local/bin/qdl', '--debug', '--storage', 'spinor', '--serial', 'FDBDE577', 'xbl_s_devprg_ns.melf', 'rawprogram0.xml', 'patch0.xml']
...
end: 2.5 flash-qdl (duration 00:05:26) [common]
```
https://lava-staging.infra.foundries.io/scheduler/job/3603#L602
After:
```
start: 2.5 flash-qdl (timeout 00:14:35) [common]
Calling: ['/usr/local/bin/qdl', '--skipblock=sha256', '--debug', '--storage', 'spinor', '--serial', 'FDBDE577', 'xbl_s_devprg_ns.melf', 'rawprogram0.xml', 'patch0.xml']
...
end: 2.5 flash-qdl (duration 00:00:09) [common]
```
https://lava-staging.infra.foundries.io/scheduler/job/3611#L630

qdl.c
```diff
@@ -1081,6 +1081,31 @@ static int qdl_flash(int argc, char **argv)
 	if (ret < 0)
 		goto out_cleanup;
 
+	/*
+	 * firehose.c only fills the digest on the op; the `sha256` subcommand
+	 * is a qdl-level concern, so emit the formatted hex here.
+	 */
+	{
```
quic-bjorande: Let's break this out to its own function.

qdl.c
```diff
@@ -1081,6 +1081,31 @@ static int qdl_flash(int argc, char **argv)
...
+		struct firehose_op *op;
+
+		list_for_each_entry(op, &firehose_ops, node) {
+			char hex[SHA256_DIGEST_STRING_LENGTH];
+			size_t i;
+
+			if (op->type != FIREHOSE_OP_GET_SHA256_DIGEST ||
+			    !op->digest_valid)
```
quic-bjorande: We asked for a digest but didn't get one. I think we should inform the user that we didn't get a digest for this one. What do you think?

firehose.c
```diff
@@ -651,6 +653,76 @@ static int firehose_program(struct qdl_device *qdl, struct firehose_op *program)
...
+	 * they match, skip the <program> entirely.
+	 *
+	 * Restricted to non-sparse, non-NAND programs with VIP disabled:
+	 *   - Sparse chunks would each need their own digest; v1 keeps them on
+	 *     the normal program path.
+	 *   - NAND has spare/OOB bytes whose semantics differ across
+	 *     programmers.
+	 *   - <getsha256digest> is not part of pre-built VIP digest tables.
+	 */
+	if (qdl_skipblock_mode == QDL_SKIPBLOCK_SHA256 &&
```
quic-bjorande: Let's try to keep firehose_program() "reasonable" in length, by breaking this out to a separate function. Something like qdl_should_skipblock(), and upon returning positively we skip the rest of firehose_program().

firehose.c
```diff
@@ -651,6 +653,76 @@ static int firehose_program(struct qdl_device *qdl, struct firehose_op *program)
...
+			if ((size_t)hn < want) {
+				size_t pad = want - hn;
+
+				memset(buf, 0, pad);
+				SHA256Update(&ctx, buf, pad);
+			}
+			hashed += want;
+		}
+
+		if (hashed == total) {
```
quic-bjorande: In what case would hashed != total? Only the early return in the middle of the loop, right?

So, if we extract this into its own function, we can put an early return in the case of read failure and this conditional should go away.

firehose.c
```diff
@@ -651,6 +653,76 @@ static int firehose_program(struct qdl_device *qdl, struct firehose_op *program)
...
+			size_t want = MIN(qdl->max_payload_size, total - hashed);
+
+			hn = qdl_file_read(&file, buf, want);
+			if (hn < 0) {
+				hashed = SIZE_MAX;
+				break;
+			}
+			if (hn > 0)
+				SHA256Update(&ctx, buf, hn);
+			if ((size_t)hn < want) {
```
quic-bjorande: read() is defined such that it is okay to return a partial read, so I think we shouldn't pad the remainder of the sector with 0s when that happens. Empirically this hasn't been a problem so far, but we should correct this before it becomes a problem...

And yes, we have the same potential issue below, in the programming loop. I think we should introduce a "qdl_file_read_exact()" wrapper that does loop until the requested amount has been read, and only return partial size if we hit EOF.

qdl.c
```diff
@@ -488,6 +489,8 @@ static void print_usage(FILE *out)
 	fprintf(out, " -D, --vip-table-path=T\t\tUse digest tables in the T folder for VIP\n");
 	fprintf(out, " -R, --skip-reset\t\tDo not send the reset command after flashing completes\n");
 	fprintf(out, "     --backend=B\t\tSelect device backend B: <auto|usb|qud> (default: auto)\n");
+	fprintf(out, "     --skipblock=M\t\tSkip <program> when on-device SHA-256 already matches;\n");
```
quic-bjorande: Perhaps something like: "Use requested readback mechanism to determine if program entries can be skipped."

qdl.c
```diff
@@ -49,6 +49,7 @@ enum {
 };
 
 bool qdl_debug;
+enum qdl_skipblock_mode qdl_skipblock_mode = QDL_SKIPBLOCK_NONE;
```
quic-bjorande: Wouldn't it be cleaner to pass this through the qdl_device struct? (Perhaps even through the firehose_ops, but they can come from a few different places, so that's more annoying to inject).

qdl.c
```diff
@@ -907,6 +912,14 @@ static int qdl_flash(int argc, char **argv)
 			    decode_backend(optarg, &qdl_dev_type) < 0)
 				errx(1, "unknown backend \"%s\" (expected auto|usb|qud)", optarg);
 			break;
+		case OPT_SKIPBLOCK:
+			if (!strcmp(optarg, "none"))
+				qdl_skipblock_mode = QDL_SKIPBLOCK_NONE;
+			else if (!strcmp(optarg, "sha256"))
+				qdl_skipblock_mode = QDL_SKIPBLOCK_SHA256;
+			else
+				errx(1, "unknown --skipblock mode \"%s\"", optarg);
```
quic-bjorande: ...", valid options are none and sha256"

quic-bjorande (commented):
Just some smaller things. Feature looks really good, nicely done!

qdl.c
igoropaniuk: moved to a dedicated function

qdl.c
igoropaniuk: add ux_err() for this case

firehose.c
igoropaniuk: done, thanks for the suggestion!

firehose.c
igoropaniuk: done, thanks!

firehose.c
igoropaniuk: Introduced qdl_file_read_exact()

qdl.c
igoropaniuk: Fixed, tjhanks!

qdl.c
igoropaniuk: done!

igoropaniuk:
@andersson @quic-bjorande all comments addressed, thanks for the review!

igoropaniuk:
Added description of the feature to the README.md (in a separate commit)

PR #243, opened 2026-05-15
sahara: auto-assemble minidump ELF after ramdump download

mukeshojha-linux:
When the kernel minidump backend is active, the Sahara debug table contains a KELF region — a complete ELF core header describing the physical address and output offset of every other kernel region.

Detect KELF after download and stitch all kernel regions into a single minidump.elf, ready for crash or gdb, without any manual concatenation. Devices without minidump enabled are unaffected.

On Windows, inline minimal ELF64 definitions since <elf.h> is absent in MinGW environments.

based on suggestion given here for meminspect patches

https://lore.kernel.org/lkml/abtlUQqMOxj5PwGB@baldur/

### DIFF
--- a/sahara.c
+++ b/sahara.c
@@ -18,6 +18,47 @@
 #include "qdl.h"
 #include "oscompat.h"
 
+/* Minimal ELF64 definitions — <elf.h> is not available on all platforms */
+#define ELFMAG		"\177ELF"
+#define SELFMAG		4
+#define ET_CORE		4
+#define PT_LOAD		1
+
+#define EI_NIDENT	16
+typedef uint16_t Elf64_Half;
+typedef uint32_t Elf64_Word;
+typedef uint64_t Elf64_Addr;
+typedef uint64_t Elf64_Off;
+typedef uint64_t Elf64_Xword;
+
+typedef struct {
+	unsigned char	e_ident[EI_NIDENT];
+	Elf64_Half	e_type;
+	Elf64_Half	e_machine;
+	Elf64_Word	e_version;
+	Elf64_Addr	e_entry;
+	Elf64_Off	e_phoff;
+	Elf64_Off	e_shoff;
+	Elf64_Word	e_flags;
+	Elf64_Half	e_ehsize;
+	Elf64_Half	e_phentsize;
+	Elf64_Half	e_phnum;
+	Elf64_Half	e_shentsize;
+	Elf64_Half	e_shnum;
+	Elf64_Half	e_shstrndx;
+} Elf64_Ehdr;
+
+typedef struct {
+	Elf64_Word	p_type;
+	Elf64_Word	p_flags;
+	Elf64_Off	p_offset;
+	Elf64_Addr	p_vaddr;
+	Elf64_Addr	p_paddr;
+	Elf64_Xword	p_filesz;
+	Elf64_Xword	p_memsz;
+	Elf64_Xword	p_align;
+} Elf64_Phdr;
+
 #define SAHARA_HELLO_CMD		0x1  /* Min protocol version 1.0 */
 #define SAHARA_HELLO_RESP_CMD		0x2  /* Min protocol version 1.0 */
 #define SAHARA_READ_DATA_CMD		0x3  /* Min protocol version 1.0 */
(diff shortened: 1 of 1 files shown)


sahara.c
```diff
@@ -18,6 +18,51 @@
 #include "qdl.h"
 #include "oscompat.h"
 
+#ifndef _WIN32
```
igoropaniuk:
```
#if defined(__linux__)
...
#endif
```

and this will address MacOS build failure as well

sahara.c
```diff
@@ -412,6 +457,208 @@ static bool sahara_debug64_filter(const char *filename, const char *filter)
...
+	/* Read the KELF file into memory so we can parse its program headers */
+	snprintf(path, sizeof(path), "%s/%s", ramdump_path, table[kelf_idx].filename);
+	in = fopen(path, "rb");
+	if (!in) {
+		ux_err("minidump: failed to open KELF file %s\n", path);
+		return;
+	}
+
+	fseek(in, 0, SEEK_END);
+	kelf_size = ftell(in);
```
igoropaniuk: please check ftell() result for -1L

sahara.c
```diff
@@ -412,6 +457,208 @@ static bool sahara_debug64_filter(const char *filename, const char *filter)
...
+				fwrite(copy_buf, 1, n, out);
+				written += n;
+			}
+			fclose(in);
+
+			/* Zero-pad to p_filesz to keep offsets aligned */
+			while (written < phdrs[j].p_filesz) {
+				char zero = 0;
+
+				fwrite(&zero, 1, 1, out);
```
igoropaniuk: writes one byte at a time, maybe we better use chunked zero buffer for this instead, e.g. :

```
static const char zeros[4096] = { 0 };
size_t n = phdrs[j].p_filesz;

while (n > 0) {
   size_t chunk = n > sizeof(zeros) ? sizeof(zeros) : n;
   if (fwrite(zeros, 1, chunk, out) != chunk)
        return -1;
   n -= chunk;
}
```

or just use sparse hole, e.g. fseek() past the gap and write one sentinel byte

I'd also move this to a distinct function

sahara.c
```diff
@@ -412,6 +457,208 @@ static bool sahara_debug64_filter(const char *filename, const char *filter)
...
+	}
+
+	/* 1. Write KELF: provides the ELF header, phdrs, and partial PT_NOTE */
+	if (fwrite(kelf_buf, 1, kelf_size, out) != kelf_size) {
+		ux_err("minidump: failed to write KELF to output\n");
+		goto out_close;
+	}
+
+	/* 2. Write vmcoreinfo immediately after KELF to complete the PT_NOTE */
+	if (vmcore_idx >= 0) {
```
igoropaniuk, line 631: I'd move this logic for fopen() + chunked read/write + fclose() to a separate function (`copy_file_to(out, path)`), so we can avoid similar functionality duplication on lines 620-635

sahara.c
```diff
@@ -412,6 +457,208 @@ static bool sahara_debug64_filter(const char *filename, const char *filter)
...
+	}
+
+	if (fread(kelf_buf, 1, kelf_size, in) != kelf_size) {
+		ux_err("minidump: failed to read KELF file\n");
+		fclose(in);
+		goto out_free;
+	}
+	fclose(in);
+
+	ehdr = kelf_buf;
```
igoropaniuk, line 579: please add sanity checks (in case ELF header is malformed) for ehdr:
```
ehdr->e_phoff > kelf_size || (size_t)ehdr->e_phnum * sizeof(Elf64_Phdr) > kelf_size - ehdr->e_phoff
```

sahara.c
quic-bjorande: The purpose of the include is to avoid duplicating definitions in the source file, as we already duplicated them below for the other OSs we could simply skip the #if and #include and just run off the same definition (unless there's some convenient way to get elf.h on Windows and MacOS)

sahara.c
```diff
@@ -412,6 +457,208 @@ static bool sahara_debug64_filter(const char *filename, const char *filter)
...
+ *
+ * This function detects those two special regions, then iterates the PT_LOAD
+ * headers in p_offset order, matches each one to a downloaded region file by
+ * physical address, and concatenates everything into a single minidump.elf
+ * ready for crash or gdb.
+ *
+ * All segments are written sequentially because meminspect assigns p_offset
+ * values in a strictly increasing, contiguous order.
+ */
+static void sahara_build_minidump_elf(const char *ramdump_path, const char *filter,
```
quic-bjorande, line 513: I first asked that you'd move this whole thing to a minidump.c that you call once sahara_run() returns. But I missed the fact that you're passing the sahara_debug_region64 table.

I still think it would be nice to split this out, as it does not relate to the sahara protocol directly. But I think it's reasonable to just call it from sahara.c directly, at least not very useful to try to get the table out to the caller at this time.

sahara.c
```diff
@@ -412,6 +457,208 @@ static bool sahara_debug64_filter(const char *filename, const char *filter)
...
+	 * Identify the vmcoreinfo region: among all K-prefixed regions, it is
+	 * the one whose physical address has no matching PT_LOAD p_paddr entry.
+	 * It fills the tail of the PT_NOTE section and must be placed immediately
+	 * after the KELF content in the output file.
+	 */
+	for (i = 0; i < n_regions; i++) {
+		if ((ssize_t)i == kelf_idx || strncmp(table[i].region, "md_K", 4) != 0)
+			continue;
+
+		if (sahara_debug64_filter(table[i].filename, filter))
```
quic-bjorande: I guess this would make sense if I grab a minidump into a directory and then again grab another minidump with stricter filter?

Please add a comment about ensuring that stale files aren't picked up.

sahara.c
mukeshojha-linux: Sure.

sahara.c
mukeshojha-linux: yeah, right.

sahara.c
mukeshojha-linux: thank you.

sahara.c
mukeshojha-linux: It makes sense.

sahara.c
mukeshojha-linux: Not moving to a separate file..

sahara.c
mukeshojha-linux: Sure, will apply.

sahara.c
mukeshojha-linux: we can drop this.,

PR #250, opened 2026-05-25
tree: move sources under src/ and consolidate on a single qdl binary

igoropaniuk:
Reorganizes the source tree and folds the standalone helper binaries into the main qdl binary.


igoropaniuk:
@andersson could you please take a look and provide some initial feedback?

andersson:
I think the moving of source into src/ makes sense, and looks clean. I'm not convinced on the second level split though, it adds the additional step of figuring out which category to find/add files in.

Concretely, I think we should drop qdl-ramdump and qdl-ks - which would result in cli being a single file. I have a patch that splits out the programmer-related functions to a programmer.c, but I don't think that's "package" and it's not "util", so is that now a new category? Is "flash" the right name for that category?

igoropaniuk:
@andersson I'll re-work and keep only one level (moving sources to `src`), and in the second commit will drop both qdl-ramdump and qdl-ks binaries, let me know if it works for you

igoropaniuk:
@andersson applied changes as you've suggested, thanks!

lumag (approved):

PR #262, opened 2026-06-16
contents: Don't ignore ignored entries

quic-bjorande:
While it would make sense to ignore entries explicitly marked to be ignored, some builds apparently depend on these ignored files.

Drop the check for ignored to make it possible to flash those builds.

### DIFF
--- a/contents.c
+++ b/contents.c
@@ -112,22 +112,6 @@ out_free_str:
 	return ret;
 }
 
-static bool contents_entry_is_ignored(xmlNode *node)
-{
-	xmlChar *ignore;
-	bool ret;
-
-	ignore = xmlGetProp(node, (xmlChar *)"ignore");
-	if (!ignore)
-		return false;
-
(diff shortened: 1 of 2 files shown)


quic-bjorande:
It's unclear if this patch is correct...

contents.c
```diff
@@ -335,9 +319,6 @@ static int contents_parse_entry(struct contents *contents, xmlNode *node,
 		return -1;
 	}
 
-	if (contents_entry_is_ignored(node))
```
lumag, line 338: I'd guess it needs some explanation, why we should not ignore the files flagged to be ignored.

andersson: I've not found any description of what it is that the "ignored" attribute implies that we should ignore, and so far it's only one target that I've run into where we need to ignore the ignore-flag.

I don't think we should merge this until we get clarity in the purpose of this attribute.

PR #267, opened 2026-06-19
README: describe how to use QDL from WSL2 on Windows

brgl:
Add a section on how to set up the USB forwarding to WSL on Windows machines and how to use qdl for flashing from the linux shell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,58 @@ or those with locked bootloaders.
 
 Please consult your device’s documentation for instructions on how to enter EDL mode.
 
+### Flashing from WSL2 (usbipd-win)
+
+On Windows, QDL can run inside a WSL2 distribution, but WSL2 does not see USB
+devices by default — they must be forwarded from the Windows host with
+[usbipd-win](https://github.com/dorssel/usbipd-win). Install it on the Windows
+side (`winget install usbipd`), then forward the EDL device into WSL.
+
+EDL devices enumerate under Vendor ID `05c6` with Product ID `9008` (Firehose),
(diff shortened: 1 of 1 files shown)


brgl:
Hi @andersson !

Sorry, I didn't pull out this PR in time. I commented on it in private but it probably didn't reach you.

There's an issue with usbipd that probably makes this useless and we should revert this PR.

The same qdl command that works fine on Windows, often fails under WSL2:

```
ERROR: Read non multiple sector size value from USB 827567 sectsize4096
FIREHOSE RAW BINARY WRITE: sailsw1.elf, 827392 bytes
```

827392 = 808 * 1024 (sector-aligned), and 827392 + 175 is exactly the length of the next <program> command: the data chunk and the following command arrive merged into a single bulk read on the device, so it's no longer sector-aligned and the write is rejected.

The trigger is the terminating ZLP. Sector-aligned payloads are always a multiple of the 1024-byte wMaxPacketSize, so they need a trailing zero-length packet to mark end-of-transfer, and that ZLP gets dropped somewhere in the USB/IP path.

The WSL-side dmesg shows the transfers being aborted:

```
vhci_hcd: unlink->seqnum 37404
vhci_hcd: urb->status -104
```

This matches a set of open usbipd-win issues, and the maintainer diagnosed it as a USB/IP latency/timing problem (transfers stalling and getting unlinked by WSL kernel after 3ms):

https://github.com/dorssel/usbipd-win/issues/924 
https://github.com/dorssel/usbipd-win/issues/1022
https://github.com/dorssel/usbipd-win/issues/1067

I think at most we should keep this description in README but add a big fat warning that this is unreliable.

PR #271, opened 2026-06-21
tests: add a hardware-in-the-loop (HIL) test suite

igoropaniuk:
Before tagging a release we want to exercise qdl against real hardware - flashing a full build, reading partitions, skipblock, sha256, erase and write - but there is no automated way to do that today. Add a "hil" meson suite that drives the built qdl binary against an attached EDL device, so a release can be validated with a single command. It is a stop-gap that simplifies testing until we have proper CI HIL coverage via LAVA.

Each step is its own test in the "hil" suite. The steps are stateful, so they run serially and in order (is_parallel:false plus a descending priority): flash the build, read the EFI partition / GPT / address forms, re-flash with --skipblock=sha256, run the sha256 command, erase and write the EFI partition back, and finally reboot. Every step runs with --skip-reset except the reboot. The erase/write pair round-trips the partition's own readback, so nothing is left wiped, and once one step fails the rest of the suite stops.

The suite is opt-in: the build and storage type are passed via the environment, and when unset every step exits 77 (SKIP), so a plain "meson test" / CI run stays green without hardware.

Usage:
```bash
  $ QDL_HIL_BUILD=/path/to/build QDL_HIL_STORAGE=ufs \
        meson test -C build --suite hil -v

   1/10 hil - qdl:HIL flash whole build       OK   21.63s
   2/10 hil - qdl:HIL read EFI partition      OK   10.50s
   3/10 hil - qdl:HIL read GPT table          OK    2.47s
   4/10 hil - qdl:HIL read address P/S+L      OK    2.46s
   5/10 hil - qdl:HIL read address P/name     OK   11.17s
   6/10 hil - qdl:HIL flash skipblock sha256  OK   22.12s
   7/10 hil - qdl:HIL sha256 command          OK    2.47s
   8/10 hil - qdl:HIL erase EFI partition     OK    4.97s
   9/10 hil - qdl:HIL write EFI partition     OK   10.18s
  10/10 hil - qdl:HIL reboot board            OK    3.64s

  Ok: 10   Fail: 0
 ```

### DIFF
--- a/tests/meson.build
+++ b/tests/meson.build
@@ -23,6 +23,45 @@ test(
   protocol: 'tap',
 )
 
+# --- hardware-in-the-loop tests ---
+# Each step is a separate test in the "hil" suite. Run explicitly against an
+# attached EDL device with:
+#   QDL_HIL_BUILD=... QDL_HIL_STORAGE=... meson test --suite hil
+# Without that environment every step exits 77 (SKIP), so the suite is
+# harmless if picked up by a plain "meson test" run.
+#
+# The steps are stateful (flash -> read -> erase -> write -> reboot), so they
+# must not overlap and must run in order: is_parallel:false serializes them
+# and a descending priority fixes the order. Flashing a full build can take a
+# while, hence the long timeout.
+hil_steps = [
+  ['flash whole build',       'flash_full',         100],
+  ['read EFI partition',      'read_partition',      90],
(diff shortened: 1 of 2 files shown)


igoropaniuk:
Tested invocation of the test suite on both Linux and Windows (msys2/mingw64 env)

PR #273, opened 2026-06-24
README: mark WSL2 support as unreliable

brgl:
Add a warning to the README saying that WSL2 support is not reliable with current usbipd-win. This may change in the future though.


brgl:
This is a follow-up to [this](https://github.com/linux-msm/qdl/pull/267#issuecomment-4767442362) comment.

PR #279, opened 2026-07-01
[RFC] doc: add AGENTS.md guide for automation agents

igoropaniuk:
Automation agents contributing to QDL need a single, machine-oriented reference for how to build, test, and format their changes. Scattering this information across the README and CONTRIBUTING.md makes it easy for agents to miss the project conventions.

Add AGENTS.md consolidating the build prerequisites, the pre-commit build and check targets, the Linux kernel coding style requirement, the subsystem-tagged commit convention with a mandatory rationale body, the ASCII-only rule for code, docs, and commit messages, and the mandatory Signed-off-by trailer. It points back to CONTRIBUTING.md so the two stay in sync.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,103 @@
+# Agent Guide for QDL
+
+This file guides automation agents working in the QDL repository. Human
+contributors should read [CONTRIBUTING.md](CONTRIBUTING.md); this document
+captures the same rules in a form suited to automated workflows.
+
+## Project overview
+
+QDL is a command-line tool for flashing firmware on Qualcomm-based devices
+over USB in Emergency Download (EDL) mode. It uploads a flash loader to the
+device and drives it using the Sahara and Firehose protocols.
+
+## Prerequisites and build
+
+### Linux
+
+```bash
+sudo apt install libxml2-dev libusb-1.0-0-dev libzip-dev meson ninja-build help2man
+meson setup build
+meson compile -C build
+```
+
+### macOS (Homebrew)
+
+```bash
+brew install libxml2 libusb libzip meson ninja help2man
+meson setup build
+meson compile -C build
+```
+
+### macOS (MacPorts)
+
+```bash
+sudo port install libxml2 libusb libzip meson ninja help2man
+meson setup build
+meson compile -C build
+```
+
+## Before committing
+
+Every change must build cleanly and pass the style and test checks. Run:
+
+```bash
+meson setup build              # once, or after changing meson.build
+meson compile -C build         # build must succeed
+meson test -C build            # run the test suite
+meson compile check -C build   # checkpatch (Linux kernel style)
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,101 @@
...
+- Wrap body lines at ~72 characters.
+- Do not mix unrelated changes in one commit; split them logically.
+- Each patch must be logically coherent, self-contained, and independently
+  buildable. The tree must remain functional after every commit.
+
+## Sign-off
+
+Every commit must carry a `Signed-off-by` trailer using the identity from the
+local git configuration. When committing programmatically, append it
+explicitly:
```
quic-kdybcio: I don't think that's right. DCO must come from a human.

igoropaniuk: I was using https://github.com/qualcomm-linux/meta-qcom/blob/master/AGENTS.md as a reference

let me know which approach may work better? thanks

quic-kdybcio: OSDO guidelines explicitly require this. meta-qcom needs to state this too (cc @ricardosalveti )

ricardosalveti: We can only suggest the agent to use the git username / email that is configured locally by the user, and then it is up for the user to make sure it is following the DCO requirements as the changes are proposed in a pull-request, not sure we can instruct the agent differently here.

quic-kdybcio: If you literally write `AI agents MUST NOT add Signed-off-by tags` like the [kernel docs](https://docs.kernel.org/process/coding-assistants.html), it will respect it (most of the time).

ricardosalveti: My thinking was to move the responsibility towards the process of opening the pull-request, but I agree it is fair to let the user add it manually (not that it will change the outcome of the result, but at least it is another step the developer will have to make). @igoropaniuk mind changing here and in meta-qcom then?

igoropaniuk: Do we also want to have Assisted-by like in Linux kernel development workflow? (https://docs.kernel.org/process/coding-assistants.html)

```
Assisted-by: AGENT_NAME:MODEL_VERSION [TOOL1] [TOOL2]
```

konradybcio: I like it. But it's not OSDO-mandatory if you have strong feelings against it

igoropaniuk: I’m not sure what we’d really gain by adding an Assisted tag. Feels like it would just add noise to the commit history with an extra line in every commit that everyone will likely ignore.

Since OSDO rules don’t require AI disclosure, I’d keep the workflow as it is. We can always adopt this rule later if needed

igoropaniuk: @konradybcio I've updated the file, added a line that you suggested

quic-kdybcio:
Let's also `ln -s AGENTS.md CLAUDE.md` because Claude ignores the former..

igoropaniuk:
@konradybcio done, thanks for the catch!

PR #280, opened 2026-07-09
sahara: add chipinfo command to read chip identity

igoropaniuk:
QDL only ever drives Sahara in its image-transfer and memory-debug
modes. Sahara also defines a command mode that lets the primary
bootloader be queried for the device's identity before any programmer
is uploaded. Expose that as a new "chipinfo" subcommand.

It enters command mode by requesting SAHARA_MODE_COMMAND in the HELLO
response and echoing the device's advertised protocol version, then
drives EXECUTE transactions to read the chip serial number, the HW ID
(broken out into MSM_ID / OEM_ID / MODEL_ID) and the OEM PK hash.
Version 3 targets no longer answer MSM_HW_ID_READ, so the same fields
are recovered through READ_CHIP_ID_V3 on those devices.

Command mode is left by switching back to image-transfer mode, which
returns the device to its power-on HELLO state and keeps it usable for
a subsequent query or flash. A Sahara reset is deliberately avoided
here: on some targets it leaves the device enumerated on PID 0x9008 but
no longer answering Sahara until it is power-cycled back into EDL.

This is handy for identifying an attached target and, in particular,
for troubleshooting secure-boot provisioning: the fused OEM PK hash and
OEM ID can be read straight from the PBL without loading a programmer.

### DIFF
--- a/src/sahara.c
+++ b/src/sahara.c
@@ -97,6 +97,14 @@ typedef struct {
 #define SAHARA_DONE_LENGTH		0x8
 #define SAHARA_DONE_RESP_LENGTH		0xc
 #define SAHARA_RESET_LENGTH		0x8
+#define SAHARA_EXECUTE_LENGTH		0xc
+#define SAHARA_SWITCH_MODE_LENGTH	0xc
+
+/* Sahara command-mode client commands, carried by SAHARA_EXECUTE_CMD */
+#define SAHARA_EXEC_CMD_SERIAL_NUM_READ		0x01
+#define SAHARA_EXEC_CMD_MSM_HW_ID_READ		0x02
+#define SAHARA_EXEC_CMD_OEM_PK_HASH_READ	0x03
+#define SAHARA_EXEC_CMD_READ_CHIP_ID_V3		0x0a
 
 #define DEBUG_BLOCK_SIZE (512u * 1024u)
 
@@ -144,6 +152,16 @@ struct sahara_pkt {
 			uint64_t offset;
 			uint64_t length;
 		} read64_req;
+		struct {
+			uint32_t client_cmd;
+		} exec_req;
+		struct {
+			uint32_t client_cmd;
+			uint32_t data_length;
+		} exec_resp;
+		struct {
+			uint32_t mode;
+		} switch_mode;
 	};
 };
 
@@ -165,13 +183,14 @@ static void sahara_send_reset(struct qdl_device *qdl)
 	qdl_write(qdl, &resp, resp.length, SAHARA_CMD_TIMEOUT_MS);
 }
 
-static int sahara_send_hello_resp(struct qdl_device *qdl, unsigned int mode)
+static int sahara_send_hello_resp(struct qdl_device *qdl, unsigned int version,
+				  unsigned int mode)
 {
 	struct sahara_pkt resp = {};
 
 	resp.cmd = SAHARA_HELLO_RESP_CMD;
 	resp.length = SAHARA_HELLO_LENGTH;
-	resp.hello_resp.version = SAHARA_VERSION;
+	resp.hello_resp.version = version;
 	resp.hello_resp.compatible = 1;
 	resp.hello_resp.status = SAHARA_SUCCESS;
 	resp.hello_resp.mode = mode;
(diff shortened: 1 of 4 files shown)


src/sahara.c
```diff
@@ -785,6 +804,308 @@ static void sahara_debug_list_images(const struct sahara_image *images)
 	}
 }
 
+static uint16_t sahara_le16(const uint8_t *p)
```
konradybcio, line 807: `#include <endian.h>`?

src/sahara.c
```diff
@@ -785,6 +804,308 @@ static void sahara_debug_list_images(const struct sahara_image *images)
...
+ * result beyond the bytes actually read.
+ */
+static size_t sahara_pkhash_trim(const uint8_t *buf, size_t len)
+{
+	static const size_t digest_sizes[] = { 32, 48, 64 };
+	size_t orig_len = len;
+	size_t i;
+
+	for (i = 4; i + 4 <= len; i++) {
+		if (!memcmp(buf, buf + i, 4)) {
```
konradybcio: I think we can afford to memcmp the entire digests - the default implementation probably bails out early on failure anyway

src/sahara.c
igoropaniuk: The `<endian.h>`is glibc-specific. macOS  has no `<endian.h>` with those macros, it exposes `OSSwapLittleToHostInt32()` via <libkern/OSByteOrder.h> instead.
Windows/MinGW  has no `<endian.h>` either.

src/sahara.c
igoropaniuk: fixed, thanks!

src/sahara.c
konradybcio: Got it, thanks

src/sahara.c
```diff
@@ -785,6 +804,308 @@ static void sahara_debug_list_images(const struct sahara_image *images)
...
+			       pkt->cmd);
+		return -1;
+	}
+
+	ret = sahara_command_info(qdl, version);
+
+	/*
+	 * Return the device to the image-transfer state it powered up in rather
+	 * than issuing a Sahara reset. A reset wedges some targets - they stay
+	 * enumerated on PID 0x9008 but stop answering Sahara - whereas a mode
```
konradybcio: A mode switch is probably preferable to keep, but isn't the issue here that the target is waiting for the host to receive some log with firehose_read?

src/sahara.c
```diff
@@ -785,6 +804,308 @@ static void sahara_debug_list_images(const struct sahara_image *images)
...
+				payload, sizeof(payload), &len) == 0 && len >= 4) {
+		serial = sahara_le32(payload);
+		have_serial = true;
+	}
+
+	if (version < 3) {
+		if (sahara_command_exec(qdl, SAHARA_EXEC_CMD_MSM_HW_ID_READ,
+					payload, sizeof(payload), &len) == 0 && len >= 8) {
+			hwid = sahara_le64(payload);
+			msm_id = (hwid >> 32) & 0xffffff;
```
konradybcio: Is it really only 24 bits?

src/sahara.c
```diff
@@ -785,6 +804,308 @@ static void sahara_debug_list_images(const struct sahara_image *images)
...
+	} else {
+		if (sahara_command_exec(qdl, SAHARA_EXEC_CMD_READ_CHIP_ID_V3,
+					payload, sizeof(payload), &len) == 0 && len >= 44) {
+			msm_id = sahara_le32(payload + 36);
+			oem_id = sahara_le16(payload + 40);
+			model_id = sahara_le16(payload + 42);
+			/* Some v3 targets carry the OEM ID in an alternate slot */
+			if (oem_id == 0 && len >= 46)
+				oem_id = sahara_le16(payload + 44);
+			hwid = ((uint64_t)(msm_id & 0xffffff) << 32) |
```
konradybcio: ditto

src/sahara.c
igoropaniuk: You're right, I shouldn't have characterized it as the reset "wedging" the target  -  I only saw the symptom. Reworded the comment to just describe the mode switch and kept it. Regarding the log point: this is the Sahara command-mode teardown (no firehose involved yet), so there's no firehose_read to drain here - the switch just returns the device to the image-transfer HELLO state.

src/sahara.c
igoropaniuk: Good catch, thanks. Dropped the 24-bit mask and treat the full 32-bit upper word as MSM_ID in both the v1/v2 and v3 paths, and widened the print to %08x

src/sahara.c
igoropaniuk: fixed

src/sahara.c
igoropaniuk: I just looked how https://github.com/bkerler/edl extracts MSM_ID (as 24 bits), in its source:

```python
self.msm_id  = int(self.hwidstr[2:8], 16)     # chars 2..7 = 6 hex = 24 bits
```
so I assumed  the top byte was reserved for something else

igoropaniuk:
@konradybcio let me know if you have other comments/suggestions. thanks!

quic-kdybcio (approved):

PR #283, opened 2026-07-13
nbdkit: expose a device LUN as a block device

igoropaniuk:
Adds an optional nbdkit plugin that uploads the firehose programmer over EDL and serves a physical partition (LUN) as a read/write block device on the host - so you can mount, dd, or image a device's live storage with ordinary tools instead of dumping partitions to files and reflashing:

```bash
modprobe nbd
nbdkit -fv ./build/nbdkit-qdl-plugin.so programmer=prog_firehose_ddr.elf lun=0
nbd-client -N default localhost 10809 /dev/nbd0
```

This revives @andersson  unmerged `nbdkit` branch. That branch predates the `meson/src/` reorg - and `master` has since independently absorbed several of its refactors (USB extraction, the JSON parser) - so it can't be cherry-picked directly. It's reconstructed here as a small adapted series on top of current master, keeping @andersson  as the author of his work.

There were ultiple adaptations & fixes added vs. the original branch
-  `pread()` reuses the chunked read path and `pwrite()` streams in max_payload_size chunks (the original did one USB transfer per sector).
- `nbdkit` calls `.open`/`.close` several times per client (`NBD_OPT_INFO`, then `NBD_OPT_GO`). Programming per open and resetting per close made the first close reboot the board and the second open time out on real hardware. The plugin now sets the device up once, keeps it configured across connections, and resets only on .unload.
- Code refactor so plugin links against the shared library.
- `getstorageinfo` parser updated to the current 3-argument `response_parser` signature and` json_parse_buf()`.


README.md
```diff
@@ -420,6 +420,96 @@ makes it possible to wire `qdl ks` into a single udev rule that covers
 multiple device configurations (for example, an optional DDR training
 image that is only present on some setups).
 
+## nbdkit plugin
```
andersson, line 423: How about moving this into a docs/nbd.md? There are a few other things in this ever growing README that would be nice to get out of it to make it easier to read.

andersson (commented):
Looks reasonable, please rebase it and I'd be happy to see this merged.

README.md
igoropaniuk: fixed, move to docs/nbd.md

PR #285, opened 2026-07-14
tests: extend HIL test suite

igoropaniuk:
Extend the hardware-in-the-loop suite with self-contained steps that fit the existing round-trip model:

- list devices: enumerate the attached EDL device with the list verb and require a vid:pid line in the output, covering device discovery over real USB.

- create-zip and flash: package a build from a contents.xml with the create-zip verb, then flash the produced zip, covering the zip writer and the zip flashing path end to end. Gated on QDL_HIL_CONTENTS.

- flash sparse build: flash a bundle that uses Android sparse images, covering the sparse decoding path against real storage. Gated on QDL_HIL_SPARSE.

- NAK on bad partition: a read of a nonexistent physical partition must fail, verifying that a device NAK is reported as an error instead of silently succeeding.

- sha256 by partition name: digest a partition addressed by its GPT name rather than by partition and sector numbers, covering GPT name resolution against the device.

- write chunked: rewrite the partition image with an overridden --out-chunk-size=16384, covering the chunked transfer path.

- verify write integrity: compare the device-computed partition digest against a local sha256 of the previously read-back and rewritten image, proving the read/write round trip is bit-exact.

The two gated steps SKIP when their environment variable is unset, and every step SKIPs without hardware, so a plain meson test stays green.

Example invocation against an attached EDL device:
```bash
  D=<path_to_build>
  QDL_HIL_BUILD=$D QDL_HIL_CONTENTS=$D/contents.xml \
    QDL_HIL_STORAGE=ufs meson test -C build --suite=hil

   1/17 hil - qdl:HIL list devices             OK          0.05s
   2/17 hil - qdl:HIL flash whole build        OK         29.03s
   3/17 hil - qdl:HIL create-zip and flash     OK         79.62s
   4/17 hil - qdl:HIL flash sparse build       SKIP        0.02s
  ...
  17/17 hil - qdl:HIL reboot board             OK          3.68s

  Ok:                16
  Fail:              0
  Skipped:           1
  ```

### DIFF
--- a/tests/test_hil.sh
+++ b/tests/test_hil.sh
@@ -25,7 +25,7 @@
 #
 # Optional environment:
 #   QDL_HIL_PROGRAMMER  Firehose programmer .elf (needed by the read/write/
-#                       erase/sha256/reboot steps). Defaults to the first
+#                       erase/sha256/reset steps). Defaults to the first
 #                       prog_firehose_*.elf next to QDL_HIL_BUILD.
 #   QDL_HIL_PARTITION   Partition for the read/erase/write steps. The EFI
 #                       system partition is small and quick to read and write,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/test_hil.sh [hunk 1]
```diff
@@ -229,26 +245,148 @@ t_write_image() {
...
+        echo "no image to write (run the read step or set QDL_HIL_IMAGE)" >&2
+        exit ${SKIP}
+    }
+    "${QEXE}" "${COMMON[@]}" -R --out-chunk-size=16384 \
+        "${PROGRAMMER}" write "${PARTITION}" "${IMAGE}"
+}
+
+# n) Integrity check: the device's digest of the partition must match a local
+# sha256 of the bytes we read back (and just wrote), proving the erase/write
+# round-trip preserved the content.
```
konradybcio, line 328: I wonder if this will be flakey on NAND, i.e. whether fh does error correction there on the fly..

konradybcio (approved):
Took a quick look only, but nothing wrong stands out

PR #286, opened 2026-07-16
Enhance crash mode device detection

ykaire-qti:
- Detect crash-mode devices by accepting USB interface protocol 0x13 (Sahara ramdump) and removing PID pre-filtering that caused ramdump devices to be silently skipped
- Add Sahara HELLO mode validation to fail early with clear errors when device mode doesn't match the operation (flash vs ramdump)

### DIFF
--- a/src/usb.c
+++ b/src/usb.c
@@ -176,10 +176,14 @@ static bool usb_match_edl_interface(const struct libusb_interface_descriptor *if
 	if (ifc->bInterfaceSubClass != 0xff)
 		return false;
 
-	/* bInterfaceProtocol of 0xff, 0x10 and 0x11 has been seen */
+	/*
+	 * Sahara is identified by bInterfaceProtocol 0x10, 0x11, or 0x13
+	 * on modern devices, and 0xff on older targets.
+	 */
 	if (ifc->bInterfaceProtocol != 0xff &&
-	    ifc->bInterfaceProtocol != 16 &&
-	    ifc->bInterfaceProtocol != 17)
+	    ifc->bInterfaceProtocol != 0x10 &&
+	    ifc->bInterfaceProtocol != 0x11 &&
+	    ifc->bInterfaceProtocol != 0x13)
 		return false;
 
 	edl->in_ep = -1;


src/usb.c
```diff
@@ -145,10 +145,16 @@ static int usb_try_open(libusb_device *dev, struct qdl_device_usb *qdl, const ch
 		if (ifc->bInterfaceSubClass != 0xff)
 			continue;
 
-		/* bInterfaceProtocol of 0xff, 0x10 and 0x11 has been seen */
+		/*
```
igoropaniuk: The comment doesn’t accurately describe all cases. I’m currently testing crashdump mode on Kodiak (QCS6490, RB3 Gen 2).

The device enters crashdump mode, but `bInterfaceProtocol` is `0x10`:
```bash
$ sudo lsusb -d 05c6: -v | grep bInterfaceProtocol
      bInterfaceProtocol     16 

```

In the meantime, QDL fetches the crash dump:
```bash
/build/qdl ramdump -o ./ramdump667
[sudo: authenticate] Password:             
Waiting for EDL device
Collecting crash dump from device (PID 0x900e, serial: 8AE002FE)
OCIMEM.BIN dumped successfully                                                   
CODERAM.BIN dumped successfully
DATARAM.BIN dumped successfully
...
```

Full `lsbusb` output:
```bash
$ sudo lsusb -d 05c6: -v
[sudo: authenticate] Password:             

Bus 007 Device 020: ID 05c6:900e Qualcomm, Inc. QUSB_BULK_SN:8AE002FE
Negotiated speed: SuperSpeed (5Gbps)
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               3.10
  bDeviceClass            0 [unknown]
  bDeviceSubClass         0 [unknown]
  bDeviceProtocol         0 
  bMaxPacketSize0         9
  idVendor           0x05c6 Qualcomm, Inc.
  idProduct          0x900e QUSB_BULK_SN:8AE002FE
  bcdDevice            0.00
  iManufacturer           1 Qualcomm CDMA Technologies MSM
  iProduct                2 QUSB_BULK_SN:8AE002FE
  iSerial                 3 f6064876
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x002c
    bNumInterfaces          1
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0xa0
      (Bus Powered)
      Remote Wakeup
    MaxPower                8mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass       255 Vendor Specific Class
      bInterfaceSubClass    255 Vendor Specific Subclass
      bInterfaceProtocol     16 
      iInterface              0 
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x81  EP 1 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0400  1x 1024 bytes
        bInterval               0
        bMaxBurst               3
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x01  EP 1 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0400  1x 1024 bytes
        bInterval               0
        bMaxBurst               3
Binary Object Store Descriptor:
  bLength                 5
  bDescriptorType        15
  wTotalLength       0x002a
  bNumDeviceCaps          3
  SuperSpeedPlus USB Device Capability:
    bLength                20
    bDescriptorType        16
    bDevCapabilityType     10
    bmAttributes         0x00000001
      Sublink Speed Attribute count 2
      Sublink Speed ID count 1
    wFunctionalitySupport   0x1100
      Min functional Speed Attribute ID: 0
      Min functional RX lanes: 1
      Min functional TX lanes: 1
    bmSublinkSpeedAttr[0]   0x000a4030
      Speed Attribute ID: 0 10Gb/s Symmetric RX SuperSpeedPlus
    bmSublinkSpeedAttr[1]   0x000a40b0
      Speed Attribute ID: 0 10Gb/s Symmetric TX SuperSpeedPlus
  SuperSpeed USB Device Capability:
    bLength                10
    bDescriptorType        16
    bDevCapabilityType      3
    bmAttributes         0x00
    wSpeedsSupported   0x000e
      Device can operate at Full Speed (12Mbps)
      Device can operate at High Speed (480Mbps)
      Device can operate at SuperSpeed (5Gbps)
    bFunctionalitySupport   1
      Lowest fully-functional device speed is Full Speed (12Mbps)
    bU1DevExitLat          10 micro seconds
    bU2DevExitLat        2047 micro seconds
  USB 2.0 Extension Device Capability:
    bLength                 7
    bDescriptorType        16
    bDevCapabilityType      2
    bmAttributes   0x00000002
      HIRD Link Power Management (LPM) Supported
Device Status:     0x000c
  (Bus Powered)
  U1 Enabled
  U2 Enabled

```

igoropaniuk (dismissed):
3ce104e (`usb: detect crash-mode devices by interface protocol`) touches only the libusb backend. IIUC it does two things there: accepts `bInterfaceProtocol 0x13` in addition to `0xff/0x10/0x11` in `usb_try_open()`, and stops pre-filtering by known EDL PIDs in `try_usb_open() `so crash-mode devices with unlisted PIDs still get counted as visible. 

The QUD backend in `src/qud.c` has no equivalent code to fix - it matches any VID_05C6 device by hardware-ID prefix and never looks at `bInterfaceProtocol` or a PID allowlist; on Windows the gate is whether the QDLoader driver binds the device, not qdl-side descriptor checks.

igoropaniuk:
FYI https://github.com/linux-msm/qdl/pull/287

src/usb.c
JohnSagaQuic: As Hang explained in the offline chat, 0x10 is a possible Crash mode case.  It can be further confirmed via the Sahara Hello.

JohnSagaQuic:
We will limit the scope of this PR for the usb.c change.   Will leave the qud.c case to Igor as he is already working on the refactoring [287](https://github.com/linux-msm/qdl/pull/287)

src/sahara.c
```diff
@@ -1171,6 +1171,37 @@ int sahara_run(struct qdl_device *qdl, const struct sahara_image *images,
 
 		switch (pkt->cmd) {
 		case SAHARA_HELLO_CMD:
+			/*
```
andersson: This change is good, but unrelated to the "crash mode device detection" problem.
Please move this code into sahara_hello() and move this commit to a separate PR.

src/usb.c
```diff
@@ -145,10 +145,16 @@ static int usb_try_open(libusb_device *dev, struct qdl_device_usb *qdl, const ch
...
+		 * bInterfaceProtocol of 0xff, 0x10 and 0x11 has been seen on
+		 * normal EDL/Firehose devices. 0x13 (DEV_PROTOCOL_SAHARA_SBL_XBL_RAMDUMP)
+		 * is used by devices in crash/ramdump mode and must also be
+		 * accepted, otherwise crash-mode devices are silently skipped.
```
andersson: In line with our internal discussion, please change this to.

"Sahara is identified by a bInterfaceProtocol of 0x10, 0x11, or 0x13 on modern devices, and 0xff on older targets."

src/usb.c
```diff
@@ -145,10 +145,16 @@ static int usb_try_open(libusb_device *dev, struct qdl_device_usb *qdl, const ch
...
+		 */
 		if (ifc->bInterfaceProtocol != 0xff &&
 		    ifc->bInterfaceProtocol != 16 &&
-		    ifc->bInterfaceProtocol != 17)
+		    ifc->bInterfaceProtocol != 17 &&
```
andersson: The mix of bases are ugly, and these numbers are given in hex format in the internal documentation. So please change 16 and 17 to base 16 as well.

src/usb.c
```diff
@@ -256,9 +262,19 @@ int try_usb_open(struct qdl_device *qdl, const char *serial, int *visible_out)
 
 		if (libusb_get_device_descriptor(dev, &desc) < 0)
 			continue;
-		if (desc.idVendor != 0x05c6 || !usb_is_edl_pid(desc.idProduct))
+		if (desc.idVendor != 0x05c6)
```
andersson: As you show in patch 3/3 this change is incomplete...

src/usb.c
```diff
@@ -256,9 +262,19 @@ int try_usb_open(struct qdl_device *qdl, const char *serial, int *visible_out)
...
 			continue;
 
+		/*
+		 * Do not pre-filter by PID here. Crash-mode devices may expose
```
andersson: It's not clear to me what purpose this comment has. It says "Do not pre-filter" and the code says "visible++;". That's not prefiltering, that is counting the number of visible devices...

Perhaps the comment is trying to document why the code doesn't call usb_is_edl_pid()? That won't be clear to future readers.

src/usb.c
```diff
@@ -26,11 +26,6 @@ struct qdl_device_usb {
 	size_t out_chunk_size;
 };
 
-static bool usb_is_edl_pid(uint16_t pid)
```
andersson: With this last change, you're doing most of the changes that I would expect in patch 1 - as you're starting to change the usb_is_edl_pid() handling.

andersson (changes requested):
Please rebase on the new changes.

Please make sure you have clear logical changes, don't do half a logical change in one commit and then the remainder of the same change in a later commit.

src/usb.c
JohnSagaQuic: ok, will do

src/usb.c
JohnSagaQuic: ok, this is more to explain why we remove the line calling !usb_is_edl_pid(desc.idProduct).  We can remove all this comments as it wont' be relative in the future.

ykaire-qti: removed comment

src/sahara.c
ykaire-qti: removed from this PR. will address in a new PR

andersson (approved):
Nice, thank you.

Note though that the actual change request was to remove the USB PID filter, so we also need a patch that removes the PID-check from qdl_is_edl_device().

igoropaniuk:
@ykaire-qti, could you please open a follow-up PR dropping PID check from `qdl_is_edl_device()`? (or let me know if you want me to create it).

@andersson after `qdl_is_edl_device()` is adjusted and PR merged I believe we're ready to create `2.8-rc1` tag

JohnSagaQuic:
Igor, you can go ahead to raise new PR for that. Also, the name of qdl_is_edl_device() might not be accurate given the logic changed, it will be only checking based on VID.

PR #287, opened 2026-07-16
[RFC] usb: decouple open/list paths and share EDL device identity across backends

igoropaniuk:
The review discussion around [#286 ](https://github.com/linux-msm/qdl/pull/286) exposed a structural problem in the libusb backend: the answer to "what does qdl consider an EDL device" was scattered across three independent call sites (the scan/wait loop, the per-device open path, and `usb_list()`), with the interface-signature check in a fourth place and the QUD backend applying no identity policy at all. Scattered copies of the same policy drift apart - [#286 ](https://github.com/linux-msm/qdl/pull/286) itself updated one call site while the others kept the old behavior.

This series consolidates the policy into single-purpose helpers, decouples matching from opening, and makes the open path read as an explicit hierarchy:

```
auto_open()                        - backend selection (libusb / QUD)
  usb_open()                       - wait loop, the backend's .open op
    usb_open_once()                - one enumeration pass per 250 ms tick
      usb_enumerate_edl_devices()  - enumeration + EDL identity policy
                                     (shared with usb_list())
      usb_open_device()            - open, match, claim one candidate
        usb_match_device()         - serial filter + EDL interface match
```

The transport-agnostic part of the identity policy (`qdl_is_edl_pid()`, `qdl_is_edl_device()`, `EDL_VID`) moves to the shared header, and the QUD backend now enforces it in its Windows enumeration as well. Future changes to device acceptance (for example a new EDL or ramdump pid, or an additional interface protocol such as the 0x13 discussed in #286) become a one-line change that every path and both backends pick up consistently.

Everything is a pure refactor except the following deliberate changes, each documented in its commit:

- `qdl list` no longer hides a device whose `iProduct` string cannot be read or whose serial exceeds the descriptor field; it is listed with (none)/truncated serial instead.
- The serial filter is evaluated once per device instead of once per matching interface, and a device whose EDL interface fails to claim is no longer retried on a hypothetical second interface.
- The QUD backend no longer treats every Qualcomm COM port as an EDL device; ports whose pid is not in the shared allowlist (for example the diag port of a phone in normal composite mode) are ignored.

### DIFF
--- a/include/qdl.h
+++ b/include/qdl.h
@@ -144,6 +144,25 @@ struct qdl_device *auto_init(void);
 int usb_open_once(struct qdl_device *qdl, const char *serial, int *visible_out);
 int qud_probe_present(void);
 
+/*
+ * EDL device identity, shared by all transport backends: Qualcomm's
+ * vendor id plus the known EDL/ramdump product ids. The libusb backend
+ * additionally requires a matching vendor-specific interface (see
+ * usb_match_edl_interface() in usb.c); on Windows the QDLoader driver
+ * performs the equivalent match before the QUD backend sees the device.
+ */
+#define QUALCOMM_VID 0x05c6
+
+static inline bool qdl_is_edl_pid(unsigned int pid)
+{
+	return pid == 0x9008 || pid == 0x900e || pid == 0x901d || pid == 0x90db;
+}
+
+static inline bool qdl_is_edl_device(unsigned int vid, unsigned int pid)
+{
+	return vid == QUALCOMM_VID && qdl_is_edl_pid(pid);
+}
+
 struct qdl_device_desc {
 	int vid;
 	int pid;
--- a/include/qdl.h
+++ b/include/qdl.h
@@ -130,7 +130,7 @@ struct qdl_device *qud_init(void);
 struct qdl_device *auto_init(void);
 
 /*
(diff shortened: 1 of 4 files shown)


igoropaniuk:
works on both Windows+QUD and Linux+libusb, on both systems full HIL test suite has succeeded

include/qdl.h
```diff
@@ -144,6 +144,25 @@ struct qdl_device *auto_init(void);
...
 int qud_probe_present(void);
 
+/*
+ * EDL device identity, shared by all transport backends: Qualcomm's
+ * vendor id plus the known EDL/ramdump product ids. The libusb backend
+ * additionally requires a matching vendor-specific interface (see
+ * usb_match_edl_interface() in usb.c); on Windows the QDLoader driver
+ * performs the equivalent match before the QUD backend sees the device.
+ */
+#define EDL_VID 0x05c6
```
quic-bjorande: That's "Qualcomm, Inc" per https://usb.org/sites/default/files/vendor_ids051920_0.pdf

So, QUALCOMM_VID would be appropriate.

quic-bjorande (changes requested):

include/qdl.h
igoropaniuk: fixed, thanks!

PR #289, opened 2026-07-17
qdl: drop the EDL product id allowlist and match the EDL interface signature when enumerating 

igoropaniuk:
The QUD backend accepts any Qualcomm COM port whose pid is on the
shared EDL allowlist, while the libusb backend identifies the EDL
function by its interface signature - vendor-specific class and
subclass with a known Sahara protocol. Windows exposes exactly that
triple without opening the device: PnP derives compatible ids of the
form "USB\Class_xx&SubClass_xx&Prot_xx" for every USB devnode,
including each interface of a composite device.

Require the same signature set the libusb backend matches (0x10, 0x11
and 0x13 on modern devices, 0xff on older targets), bringing the two
backends' notion of an EDL device in line and filtering Qualcomm COM
ports that expose no vendor-specific function at all

Also trust the vendor id alone in qdl_is_edl_device() and drop the
now-redundant pid gate in the QUD enumeration. The pid parameter
stays in the signature so call sites are unchanged; the interface
gates are what reject non-EDL Qualcomm devices from here on.

### DIFF
--- a/src/qud.c
+++ b/src/qud.c
@@ -184,6 +184,45 @@ static unsigned int win_parse_pid(const char *hwid)
 	return pid;
 }
 
+/*
+ * Windows encodes each USB interface's class/subclass/protocol triple
+ * in the devnode's compatible ids as "USB\Class_xx&SubClass_xx&Prot_xx".
+ * Match the same EDL interface signature as the libusb backend
+ * (usb_match_edl_interface() in usb.c): vendor-specific class and
+ * subclass with a known Sahara protocol - 0x10, 0x11 or 0x13 on modern
+ * devices, 0xff on older targets.
+ */
+static bool win_match_edl_interface(HDEVINFO devinfo, SP_DEVINFO_DATA *info)
+{
+	static const char * const edl_ids[] = {
+		"USB\\Class_FF&SubClass_FF&Prot_10",
+		"USB\\Class_FF&SubClass_FF&Prot_11",
+		"USB\\Class_FF&SubClass_FF&Prot_13",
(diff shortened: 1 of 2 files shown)


src/qud.c
```diff
@@ -184,6 +184,45 @@ static unsigned int win_parse_pid(const char *hwid)
...
+		"USB\\Class_FF&SubClass_FF&Prot_11",
+		"USB\\Class_FF&SubClass_FF&Prot_13",
+		"USB\\Class_FF&SubClass_FF&Prot_FF",
+	};
+	char ids[1024];
+	DWORD type = 0;
+	const char *p;
+	size_t i;
+
+	if (!SetupDiGetDeviceRegistryPropertyA(devinfo, info,
```
quic-kdybcio, line 208: The types seem to differ slightly: https://learn.microsoft.com/en-us/windows/win32/api/setupapi/nf-setupapi-setupdigetdeviceregistrypropertya

igoropaniuk: `PropertyBuffer` is `PBYTE`, the `buffer` is `char[]` - this one is fine and deliberate. 
`info` is already an `SP_DEVINFO_DATA *` parameter here, so it matches `PSP_DEVINFO_DATA` exactly
Or you meant `PropertyBufferSize` and `sizeof(ids)`, `DWORD` vs `size_t` ?

quic-kdybcio: I only skimmed the docs and noticed they're not exactly the same. I think it's generally fine.

quic-kdybcio (approved):

PR #290, opened 2026-07-17
usb: Accept 0x9060 as USB PID for ramdump

quic-bjorande:
Extend the list of accepted USB PIDs for ramdump with 0x9060.

The proper change is to remove the USB PID list, and this is being done for an upcoming release.


igoropaniuk:
@quic-bjorande what about?  https://github.com/linux-msm/qdl/blob/f540e59dbeb462242239778ae0993549304f3abb/usb.c#L279

andersson:
> @quic-bjorande what about?
> 
You're right, and I believe it would be wrong for some other ramdump targets as well then.

Let's leave it, as the change was validated in its current form and make sure that it's correct in the upcoming v2.8 release.

PR #291, opened 2026-07-18
Address oscompat warnings on Windows

igoropaniuk:
-  rmtree() concatenates the directory path and entry name into a
   PATH_MAX buffer without checking for truncation, which -Wformat-
   truncation flags on the Windows build where PATH_MAX is 260. A
   truncated path would also make the cleanup delete the wrong file in
   principle.

- The Windows branch reads path[1] guarded only by short-circuit
   evaluation, which is correct at runtime but trips -Warray-bounds when
   the function is inlined with a string literal shorter than two bytes,
   as the oscompat unit tests do with "". It also passes a plain char
   to isalpha(), which is undefined behavior for negative values.

### DIFF
--- a/tests/test_oscompat.c
+++ b/tests/test_oscompat.c
@@ -39,7 +39,13 @@ static void rmtree(const char *path)
 			if (!strcmp(ent->d_name, ".") || !strcmp(ent->d_name, ".."))
 				continue;
 
-			snprintf(child, sizeof(child), "%s/%s", path, ent->d_name);
+			if (snprintf(child, sizeof(child), "%s/%s", path,
+				     ent->d_name) >= (int)sizeof(child)) {
+				fprintf(stderr,
+					"skipping overly-long path %s/%s\n",
+					path, ent->d_name);
+				continue;
+			}
 			if (stat(child, &st) == 0 && S_ISDIR(st.st_mode))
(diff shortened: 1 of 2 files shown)


tests/test_oscompat.c
```diff
@@ -39,7 +39,9 @@ static void rmtree(const char *path)
 			if (!strcmp(ent->d_name, ".") || !strcmp(ent->d_name, ".."))
 				continue;
 
-			snprintf(child, sizeof(child), "%s/%s", path, ent->d_name);
+			if (snprintf(child, sizeof(child), "%s/%s", path,
+				     ent->d_name) >= (int)sizeof(child))
+				continue;
```
quic-kdybcio, line 47: Should we warn here?

igoropaniuk: right call, I've added the warning there

quic-kdybcio: `overly-long`

`__func__` is generally discouraged in the kernel coding style. I think we should follow. Feel free to merge with that.

quic-kdybcio (approved):

tests/test_oscompat.c
igoropaniuk: fixed, thanks!

PR #296, opened 2026-07-23
Run the HIL suite unattended on self-hosted runners

### LINKED ISSUE #259: Establish hardware-in-the-loop testing (open board + secure-boot board with VIP)
QDL is validated today entirely in software: the meson test suite drives the sim backend (`--dry-run`) plus `cmocka` unit tests, and CI runs only on GitHub-hosted Linux/macOS/Windows runners. Nothing exercises a real device over USB/EDL, so whole classes of behavior go untested -  actual Firehose programming, storage-specific paths, reset/ramdump, and most importantly VIP (Validated Image Programming) end-to-end against a secure-boot target. We should stand up periodic hardware-in-the-loop (HIL) testing using existing LAVA lab (like it's currently done in meta-qcom, f.g. https://lava-staging.infra.foundries.io/scheduler/job/4109) covering two representative targets: an open board and a secure-boot board running in VIP mode.

We can start enabling testing configuration for the existing boards in the LAVA lab (e.g. RB3Gen2).

igoropaniuk:
Until now the HIL suite assumed a human had already put the board into
EDL mode, so it could only be run by hand. This series makes the suite
fully unattended and adds a workflow that runs it nightly on self-hosted
runners with real boards attached.

The rig drives everything itself: an environment hook switches the board
into EDL at the start of a run, the suite waits for the device to
enumerate, targets it by serial on multi-board hosts, and resets it back
into its OS at the end so the next run can reach it again. Board
specifics never enter the repository - each runner describes its boards
through env files on the host, selected via repo variables or a dispatch
input. A new hil-vip suite additionally covers Validated Image
Programming end to end on secured boards (the feature that regressed
unnoticed in v2.6), and failing runs capture qdl's debug log so they ship
their own post-mortem. A few reliability fixes shaken out by unattended
operation (fail-fast ordering, serial vs verb parsing, digest case) are
included.

The workflow triggers only on workflow_dispatch and a nightly schedule -
never on pull requests, since the runners execute the checked-out tree
against physical hardware - and a concurrency group serializes sweeps.

Merging needs repo-side setup: register the runners, set
HIL_LINUX_BOARDS, and require approval for outside collaborators'
workflow runs.

### DIFF
--- a/.github/workflows/hil.yml
+++ b/.github/workflows/hil.yml
@@ -1,0 +1,128 @@
+name: HIL
+
+# Hardware-in-the-loop runs: self-hosted runners with a device in EDL
+# reach attached (for example an Arduino Uno-Q). Board specifics stay out
+# of the repository: each runner provides QDL_HIL_BUILD, QDL_HIL_STORAGE,
+# QDL_HIL_ENTER_EDL_CMD and friends through the actions-runner .env file,
+# and tests/test_hil.sh skips cleanly when they are absent.
+#
+# SECURITY: these runners execute the checked-out tree against real
+# hardware, so this workflow must never trigger on pull requests from
+# forks. Keep the trigger to on-demand dispatch only, so a sweep runs
+# when someone asks for it and a repository without HIL runners never
+# queues a job it cannot service.
+on:
+  workflow_dispatch:
+    inputs:
+      boards:
+        description: 'Linux: JSON list of board config names (~/hil/<name>.env on the runner), or ["default"] for the runner .env'
+        required: false
+        default: ''
+      windows_boards:
+        description: 'Windows: JSON list of board config names ($HOME/hil/<name>.env under MSYS2), or ["default"] for the runner .env'
+        required: false
+        default: ''
+
+# The runners drive physical boards; run one HIL sweep at a time.
+concurrency:
+  group: hil
+  cancel-in-progress: false
+
+jobs:
+  hil-unix:
+    strategy:
+      fail-fast: false
+      matrix:
+        # macos disabled until a runner exists; restore with: [linux, macos]
+        os: [linux]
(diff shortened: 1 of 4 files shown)


igoropaniuk:
I would like to get this merged into the main repo, even though the HIL runs will keep executing only in my fork for now.

The reason is purely practical: today I have to re-apply this whole series on top of my fork every time before I can dispatch the workflow, which is error-prone and slows down every test cycle. Having it live upstream means the workflow and the test scripts are the single source of truth - my fork just tracks master and I can trigger the runs directly, no manual patching.

This is safe to carry upstream because the workflow never runs on the linux-msm/qdl repo:

- It only triggers on workflow_dispatch and the nightly schedule - never on push or pull_request, so it can't be abused by fork PRs against real hardware.
- The jobs require self-hosted runners labeled `self-hosted, hil, <os>`, which do not exist upstream, so nothing executes there.

To make that explicit (and to keep upstream's Actions tab clean from perpetually-pending scheduled runs), I can add a repository guard to each job so upstream skips them immediately:

```yaml
jobs:
  hil-unix:
    if: github.repository != 'linux-msm/qdl'
    ...
  hil-windows:
    if: github.repository != 'linux-msm/qdl'
    ...
```

With that guard, the jobs are skipped instantly in linux-msm/qdl (no queued/pending runs, no consumed minutes) and run normally in my fork, where the runners are already registered. This arrangement is meant as an interim setup until we have a proper self-hosted runner farm in Cambridge. Once those runners are online with the `hil` labels, we can drop the guard (or switch it to `== linux-msm/qdl`) and let the suite run against the official hardware directly.

Let me know if you'd prefer the stricter form (`if: github.repository == 'igoropaniuk/qdl'`) so it runs only in my fork rather than any fork with runners - I'm happy either way.

PR #298, opened 2026-07-27
brew: add Homebrew formula for macOS

idlethread:
## Summary
- Add a Homebrew formula (`Formula/qdl.rb`) so macOS users can install qdl
  via `brew tap linux-msm/qdl https://github.com/linux-msm/qdl` followed
  by `brew install --HEAD linux-msm/qdl/qdl`.
- The formula supports both `--HEAD` (Meson) and the stable v2.6 release
  (GNU Make), since the build system migrated to Meson after v2.6.
- Add a CI workflow (`.github/workflows/brew.yml`) that, on each version
  tag push, downloads the release tarball, computes its SHA256, and
  updates `Formula/qdl.rb` automatically so the in-tree formula stays in
  sync with releases without manual intervention.

## Test plan
- [x] `meson setup build && meson compile -C build` succeeds on macOS
      (Homebrew libusb, libxml2, libzip, meson, ninja, help2man)
- [x] `meson compile manpages -C build && meson install -C build --destdir ...`
      installs `qdl` and `qdl.1` as the formula's HEAD install block expects
- [x] Verified the v2.6 tarball SHA256 in the formula matches the actual
      GitHub release archive
- [x] `meson compile check-range -C build` (checkpatch) passes clean on
      both commits
- [x] Commit messages pass commitlint (subsystem-prefixed subject,
      rationale body, Signed-off-by trailer)

### DIFF
--- a/.github/workflows/brew.yml
+++ b/.github/workflows/brew.yml
@@ -1,0 +1,43 @@
+name: Update Homebrew formula
+
+on:
+  push:
+    tags:
+      - 'v[0-9]*.[0-9]*'
+
+jobs:
+  update-formula:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout master
+        uses: actions/checkout@v4
+        with:
+          ref: master
+          fetch-depth: 0
+
+      - name: Compute SHA256
+        id: vars
+        run: |
+          TAG="${{ github.ref_name }}"
+          VERSION="${TAG#v}"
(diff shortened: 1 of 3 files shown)


igoropaniuk (changes requested):
Although I like the idea of adding a Homebrew formula for QDL, I’m against direct bot pushes to master and keeping the formula file inside the repostiory, 

IMO a separate `linux-msm/homebrew-qdl` tap repo is the conventional route, and it would dissolve most of the workflow problems. User could just run `brew tap linux-msm/qdl` in this case and brew will auto-resolve `homebrew-qdl` rep + no bot commits to qdl's master are needed + `brew update` will pull only formula-containing repo instead the whole QDL repo.

PR #299, opened 2026-07-28
meson: make zip container support optional

igoropaniuk:
qdl unconditionally links against libzip, although zip archives are
just one of the supported input formats: both the flash-from-zip path
and the create-zip subcommand are conveniences on top of the plain
contents.xml / flashmap.json flows.

That unconditional dependency gets in the way of integrating qdl into
minimal embedded distributions. In Yocto, for example, libzip lives in
meta-openembedded rather than oe-core, so packaging qdl means pulling
in a whole extra layer for a feature many products never use.

Introduce a zip-container feature option and gate all libzip-backed
code on it:

- move the libzip implementation out of file.c into file_zip.c, which
  is only compiled when the feature is enabled; file.c keeps the plain
  POSIX file handling and dispatches through small helpers that turn
  into inert stubs in file.h otherwise
- provide no-zip fallbacks of qdl_zip_open/get/put where open reports
  "not a zip archive", so the contents.xml and flashmap.json flows
  work unchanged; these remain real symbols rather than inline stubs
  because the unit tests replace them with mock definitions
- build zipper.c only with the feature and stub out zipper_write(),
  with create-zip and a failed flash of a zip archive reporting that
  this qdl was built without zip-container support

The feature defaults to enabled rather than auto, so a missing libzip
is a hard configure error instead of a silently skipped dependency:
distro builds cannot accidentally lose zip support, and producing a
leaner qdl requires an explicit -Dzip-container=disabled opt-out.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -13,8 +13,17 @@ lib_sources = files(
   'io.c', 'patch.c',
   'program.c', 'read.c', 'sahara_config.c', 'sha2.c', 'sim.c', 'ufs.c', 'usb.c',
   'vip.c', 'sparse.c', 'gpt.c', 'flashmap.c', 'json.c', 'contents.c', 'pathbuf.c',
-  'zipper.c',
 )
+
+# Zip container support: reading flash bundles (file_zip.c) and the
+# create-zip subcommand (zipper.c) both need libzip. Without it,
+# file_nozip.c implements the same entry points as inert stubs.
+if libzip_dep.found()
+  common_sources += files('file_zip.c')
+  lib_sources += files('zipper.c')
+else
+  common_sources += files('file_nozip.c')
+endif
 
 # qdl: the full flashing tool (shared sources plus the CLI front-end).
 qdl_sources = lib_sources + files('qdl.c')
--- a/README.md
+++ b/README.md
@@ -63,6 +63,17 @@ meson setup build
 meson compile -C build
 ```
 
+### Build options
+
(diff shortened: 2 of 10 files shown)


konradybcio:
Can we make this an explicit opt-out instead of relying on library autodetection? I fear this may be easily overlooked and lead to not-fully-featured builds landing in distros

igoropaniuk:
> Can we make this an explicit opt-out instead of relying on library autodetection? I fear this may be easily overlooked and lead to not-fully-featured builds landing in distros

Makes sense, will fix, thanks!

src/meson.build
```diff
@@ -13,9 +13,15 @@ lib_sources = files(
...
   'vip.c', 'sparse.c', 'gpt.c', 'flashmap.c', 'json.c', 'contents.c', 'pathbuf.c',
-  'zipper.c',
 )
 
+# Zip container support: reading flash bundles (file_zip.c) and the
+# create-zip subcommand (zipper.c) both need libzip.
+if libzip_dep.found()
+  common_sources += files('file_zip.c')
+  lib_sources += files('zipper.c')
+endif
```
andersson, line 26: Would it be possible to add an else clause here with a file_nozip.c, which implements the "stub" functions, and thereby replace most of the pre-processor conditionals in the code?

igoropaniuk: makes total sense! I've pushed updated version

andersson (approved):
Looks good.

PR #300, opened 2026-08-06
qdl: let kickstart use the standard device backends

igoropaniuk:
qdl ks required -p and drove the named node with raw read()/write(),
bypassing the device backends entirely. That limits the subcommand to
devices some kernel driver has already exposed as a node, and leaves
the common case unreachable: an EDL device sitting on the USB bus has
no node to name.

It rules out Windows altogether. There an EDL device is bound either to
the Qualcomm QDLoader driver, which the QUD backend speaks, or to
WinUSB, which libusb speaks; neither is reachable through a file
descriptor. Users were told to find a serial port that qdl could not
have used even if they found it.

Make -p optional. When it is omitted, allocate a backend exactly as
ramdump and chipinfo do, so --backend and --serial pick the device and
the transport is libusb or QUD. The Sahara layer only needs read and
write hooks, so both transports share it unchanged and kickstart keeps
stopping after Sahara. Reject -p combined with --serial or --backend
rather than silently ignoring options that cannot apply.

This makes qdl a drop-in replacement for QSaharaServer on boards that
boot flashlessly, where the SoC pulls each image over Sahara and no
Firehose programmer is ever involved. A sequence that needed a COM port
and a separate tool:

```bash
    QSaharaServer.exe -u COM8 -s 13:sbl1_flashless.mbn
    QSaharaServer.exe -u COM8 -s 1:cdt.bin -s 25:tz.mbn \
                              -s 34:devcfg.mbn -s 5:u-boot.mbn
```
becomes, with the device found the same way every other subcommand
finds it:

```bash
    qdl ks -s 13:sbl1_flashless.mbn
    qdl ks -s 1:cdt.bin -s 25:tz.mbn -s 34:devcfg.mbn -s 5:u-boot.mbn
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,20 +399,33 @@ images from the host to the device. It targets *flashless boot* devices
 such as the Qualcomm Cloud AI 100, which fetch their runtime firmware
 from the host on every boot rather than storing it on-device.
 
-Unlike normal flashing, kickstart does not use USB or Firehose; it
-talks to a kernel-provided device node using plain open/read/write
-operations. Its argument set is correspondingly minimal.
+Unlike normal flashing, kickstart stops once Sahara is done: no Firehose
+programmer is uploaded and nothing is written to storage.
 
-Two arguments are required: `-p` selects the Sahara port (a device node)
-and `-s id:path` registers an image mapping. The `-s` option may be
-specified more than once, one mapping per Sahara image id the device may
-request.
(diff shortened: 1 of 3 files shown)


quic-bjorande:
LGTM.

Perhaps we should omit support for -S and only allow --serial, to limit the inconsistency between the tools?

It might also make sense to extend this to allow passing sahara archives to this thing? (Separate PR)

PR #302, opened 2026-08-10
Harden error paths across parsers, backends and the build

igoropaniuk:
Second batch of fixes from the same audit that produced [#284](https://github.com/linux-msm/qdl/pull/284): MEDIUM-severity hardening of error paths, resource cleanup and input validation. No behavior
changes on the happy path.

- **json**: fix error-path leaks, reject oversized input, and accept
  subnormal numbers instead of rejecting them on ERANGE
- **contents**: check op allocation and free the consumed programmer blob
- **patch**: abort on malformed patch entries instead of flashing past them
- **read**: propagate parse failures and fix op string leaks
- **sim**: bound sim_read() to the caller's buffer
- **usb**: reject an EDL interface lacking usable bulk endpoints
  (prevents a division by zero and transfers on endpoint -1)
- **vip**: fix the signed-table fd check (`!fd` -> `fd < 0`) and make a
  disk-full during --create-digest fail cleanly instead of crashing
- **sparse**: detect seek failures on zip-backed images
- **meson**: set _FILE_OFFSET_BITS=64 project-wide instead of per-file

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -4,6 +4,12 @@ project('qdl', 'c',
   meson_version: '>=1.1.0'
 )
 
+# Use 64-bit file offsets everywhere. This must be consistent across every
+# translation unit: off_t appears in shared types (e.g. struct firehose_op)
+# and in the qdl_file_seek() ABI, so a per-file definition would give
+# mismatched struct layouts on 32-bit builds.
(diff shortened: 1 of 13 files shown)


igoropaniuk:
@andersson @konradybcio any comments/suggestions on this? thanks!

PR #303, opened 2026-08-10
For linux msm/create sahara archive

quic-bjorande:
QDL was taught to consume Sahara payload from a custom cpio-package, but no convenient tooling was provided to generate such files.

Introduce a mechanism that allow such archive to be created directly from a contents.xml file or a list of images.

### DIFF
--- a/src/qdl.c
+++ b/src/qdl.c
@@ -10,6 +10,7 @@
 #include <getopt.h>
 #include <libgen.h>
 #include <stdbool.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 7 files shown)


src/qdl.c
```diff
@@ -348,6 +349,118 @@ static int decode_programmer(char *s, struct sahara_image *images)
...
+			snprintf(name, sizeof(name), "%d", i);
+
+		if (sahara_archive_write_entry(out, ino++, name, images[i].ptr,
+					       images[i].len) < 0)
+			goto out_close;
+	}
+
+	ret = sahara_archive_write_entry(out, ino++, "TRAILER!!!", NULL, 0);
+
+out_close:
```
igoropaniuk, line 454: A failed write leaves a truncated archive behind,  should we also `remove(filename)`  in case of error?

src/qdl.c
```diff
@@ -960,6 +1077,81 @@ static int qdl_create_zip(int argc, char **argv)
...
+	archive = argv[1];
+	filename = qdl_split_specifier(argv[2], &specifier);
+	if (!filename) {
+		ux_err("failed to parse Sahara archive input \"%s\"\n", argv[2]);
+		ret = -1;
+		goto out_free_images;
+	}
+
+	if (qdl_is_contents_xml(filename)) {
+		ret = contents_load_programmers(filename, specifier, images);
```
igoropaniuk, line 1124: `contents_load_programmers()` decodes the selector (so `contents.xml::ufs` parses fine and errors are reported), but seems `contents_find_programmers()` never sees it - it walks the entry list unfiltered and returns the first programmer XML it finds. So `::ufs` and `::emmc` produce byte-identical archives with no hint that the selector was ignored.

igoropaniuk:
two comments, none blocking, apart of this LGTM

src/qdl.c
```diff
@@ -363,6 +476,9 @@ static void print_usage(FILE *out)
 	fprintf(out, "       %s ks [-p <sahara-dev-node> | --serial=T] -s <id:file-path>...\n", __progname);
 	fprintf(out, "       %s flash (<flashmap>[::specifier] | <contents>[::<specifier>])\n", __progname);
 	fprintf(out, "       %s create-zip <zipfile> <contents>[::<specifier>]\n", __progname);
+	fprintf(out, "       %s sahara-archive <archive.bin> "
```
igoropaniuk: `create-sahara-archive` maybe? so it aligns with `create-zip` naming

igoropaniuk:
LGTM

PR #305, opened 2026-08-14
Accept NAND-style erase tags without physical_partition_number

igoropaniuk:
A user flashing a Telit FE910C04 (SDX35, NAND) reported that qdl rejects
the vendor-shipped rawprogram files with "errors while parsing erase
tag". The autogenerated NAND bundles omit `physical_partition_number`
from <erase> tags - NAND exposes a single physical partition - but
`load_erase_tag()` treated the attribute as required. The reference
`fh_loader` treats it as optional and defaults it to 0, so qdl was
stricter than the format it implements.

- program: treat `physical_partition_number` as optional in erase tags,
  defaulting to 0 to match the reference implementation. Covered by two
  new cmocka cases in `test_program_load_xml` (attribute omitted ->
  partition 0; attribute present -> honored).

- qdl: recognize erase-only XML files as program files. Found while
  verifying the above: the file type detector classified <data>
  documents by their program/read/ufs children only, so a rawprogram
  containing nothing but erase tags - as generated for NAND wipe
  flows - failed with "failed to detect file type" even though the
  parser fully supports erase tags. The flashmap loader already
  accepts erase children; do the same here.

Verified against the reported Telit rawprogram: parsing now succeeds
and qdl proceeds to device discovery.

### DIFF
--- a/src/program.c
+++ b/src/program.c
@@ -38,11 +38,19 @@ static int load_erase_tag(struct list_head *ops, xmlNode *node, bool is_nand)
 
 	program->sector_size = attr_as_unsigned(node, "SECTOR_SIZE_IN_BYTES", &errors);
 	program->num_sectors = attr_as_unsigned(node, "num_partition_sectors", &errors);
-	program->partition = attr_as_unsigned(node, "physical_partition_number", &errors);
 	program->start_sector = attr_as_string(node, "start_sector", &errors);
 	if (is_nand) {
 		program->pages_per_block = attr_as_unsigned(node, "PAGES_PER_BLOCK", &errors);
 	}
+
+	/*
+	 * NAND-generated rawprogram files omit physical_partition_number,
+	 * as NAND exposes a single physical partition. Treat the attribute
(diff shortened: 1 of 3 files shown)


src/program.c
```diff
@@ -38,12 +38,20 @@ static int load_erase_tag(struct list_head *ops, xmlNode *node, bool is_nand)
...
 		program->pages_per_block = attr_as_unsigned(node, "PAGES_PER_BLOCK", &errors);
 	}
 
+	/*
+	 * NAND-generated rawprogram files omit physical_partition_number,
+	 * as NAND exposes a single physical partition. Treat the attribute
+	 * as optional and default to partition 0, matching the reference
+	 * fh_loader behavior.
+	 */
+	if (xmlHasProp(node, (xmlChar *)"physical_partition_number"))
```
quic-kdybcio, line 52: I know this is stricter than the reference implementation, but I'd still do `if (storage != nand) { warn() }`, to reduce potential for a nasty debugging session

PR #306, opened 2026-08-15
firehose: send the partition label in NAND program tags

franciscojsfc:
Some NAND programmers resolve the destination partition by name in order to apply its bad-block adjusted offset, so start-sector alone does not describe where the data lands.
Given no label such a programmer matches no partition and programming fails:

	LOG: INFO: Program,Partition label is
	LOG: INFO: Partition need to be adjusted
	LOG: INFO: partitionTableAdusted[pi].name is
	LOG: INFO: partitionTableAdjusted[pi].length is 0
	LOG: INFO: [1122] out of good blocks, total blocks:0xa badblocks:0x0 erased blocks:0xa
	LOG: ERROR: [2052] error : 9
	LOG: ERROR Write Failed sector 0, size 0 result 2

Forward the label that the rawprogram XML already provides and flashing works.

Tested on a device which failed programming when missing this patch.
Tested on another device which worked correctly prior to this change, and keeps working
correctly after this modification.
Programming on both had the label present in rawprogram_nand.xml.

### DIFF
--- a/src/firehose.c
+++ b/src/firehose.c
@@ -1032,6 +1032,16 @@ static int firehose_program(struct qdl_device *qdl, struct firehose_op *program)
 	if (program->is_nand) {
 		xml_setpropf(node, "PAGES_PER_BLOCK", "%d", program->pages_per_block);
 		xml_setpropf(node, "last_sector", "%d", program->last_sector);
+		/*
+		 * Some NAND programmers resolve the destination partition by
+		 * name to apply that partition's bad-block adjusted offset, so
+		 * start_sector alone does not describe where the data lands.
+		 * Given no label such a programmer matches no partition and
+		 * programming fails. Programmers that do not need the name
+		 * ignore the attribute.
+		 */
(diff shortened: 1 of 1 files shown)


igoropaniuk:
I've checked reference implementation in fh_loader, and `is_nand` gate is narrower than reference behavior. `fh_loader` forwards label for every storage type, not just NAND. Gating on NAND is a defensible minimal-blast-radius choice for this fix, but it means qdl still diverges from the reference tool on UFS/eMMC storages - if a programmer ever resolves by name there too, we'll probably be back with the same issue.

Same logic applies to `<erase>` on these programmers. A NAND programmer that resolves partitions by name for programming will presumably want the name for erasing too. qdl's load_erase_tag doesn't even parse label today, so erase tags' labels (where vendors provide them) are dropped earlier in the pipeline. This is a candidate follow-up, not part of this PR.

@andersson @konradybcio I would like to have your opinion on this change.

Issue #304, opened 2026-08-12
Does QDL support rebooting a device from crashed mode?


tmakumire:
This is a feature that I often use when using working remotely with our
QCS8550 based SBC. 
When the SoC goes into crashed mode, I would just do:
PCAT -MODE RESET -DEVICE <ID>

The board would then reboot. Is there an equivalent feature in QDL?

igoropaniuk:
Hi @tmakumire ,

have you tried?:
```
qdl <firehose_ddr_binary> reset
```

igoropaniuk:
the previous suggestion when firehose programmer is loaded, which is not the case in ramdump/crash mode.

Please try https://github.com/linux-msm/qdl/pull/307

tmakumire:
Thanks @igoropaniuk, I'll try it out tomorrow

igoropaniuk:
@tmakumire did you get a chance to try it out?

tmakumire:
Hi @igoropaniuk , I just tested it and it worked great! Thank you very much.

I think that's probably one of last things I needed before I fully move away from PCAT.
I'll also check if flashing by contents.xml is working since I saw that got merged not too long ago.

PR #307, opened 2026-08-17
qdl: add a standalone reset subcommand

igoropaniuk:
Resetting a device today requires the full flashing flow: `qdl <prog.mbn> reset` uploads the programmer over Sahara just to send a Firehose power reset. That is a poor fit for the common case of kicking a board out of EDL or crash mode remotely, where no programmer for the target may even be at hand.
    
Add `qdl reset`, which resets the device from whichever state it is in: over Sahara when the device is in EDL or crash mode, falling back to a Firehose power reset when a programmer is already running (for example after a --skip-reset invocation). The existing "reset" flashing verb is unchanged.
    
Requested in issue #304 for recovering remote boards from crash mode.
    
[1] https://github.com/linux-msm/qdl/issues/304

### DIFF
--- a/src/qdl.c
+++ b/src/qdl.c
@@ -438,6 +438,88 @@ enum {
 	OPT_SERIAL,
 };
 
+/* Results of qdl_common_opt() for options every subcommand shares. */
+enum {
+	QDL_OPT_HANDLED,
+	QDL_OPT_EXIT_OK,
+	QDL_OPT_EXIT_FAIL,
+};
+
+/* Long options for subcommands that take no extra options. */
+static const struct option qdl_common_options[] = {
+	{"debug", no_argument, 0, 'd'},
+	{"version", no_argument, 0, 'v'},
+	{"serial", required_argument, 0, 'S'},
+	{"backend", required_argument, 0, OPT_BACKEND},
+	{"help", no_argument, 0, 'h'},
+	{0, 0, 0, 0}
+};
+
+/*
+ * Handle the option cases shared by every subcommand: --debug,
+ * --version, --serial, --backend and --help, plus the unknown-option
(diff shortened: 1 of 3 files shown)


src/sahara.c
```diff
@@ -1131,6 +1133,149 @@ int sahara_chipinfo(struct qdl_device *qdl)
...
+	int i;
+
+	if (qdl->dev_type == QDL_DEVICE_SIM)
+		return 0;
+
+	n = qdl_read(qdl, buf, sizeof(buf), SAHARA_CMD_TIMEOUT_MS);
+
+	/* A Firehose programmer is already running; there is no Sahara here */
+	if (n >= 5 && !memcmp(buf, "<?xml", 5))
+		return 1;
```
quic-kdybcio: This case should have an explicit message to the user, I think

src/qdl.c
```diff
@@ -620,6 +621,27 @@ static int qdl_sahara_cmd(int argc, char **argv, int (*run)(struct qdl_device *q
...
+ * Firehose power reset when a programmer is already running. Unlike the
+ * "reset" flashing verb, no programmer needs to be uploaded.
+ */
+static int qdl_reset_run(struct qdl_device *qdl)
+{
+	int ret;
+
+	ret = sahara_device_reset(qdl);
+	if (ret == 1) {
+		ux_info("Firehose programmer is running; resetting via Firehose\n");
```
quic-kdybcio: Oh it does.. but only on the caller side. Let's move it to where it's determined

src/sahara.c
igoropaniuk: fixed

src/qdl.c
igoropaniuk: fixed

quic-kdybcio (approved):

PR #309, opened 2026-08-25
ufs: replace module globals with a caller-owned context

igoropaniuk:
The UFS provisioning loader kept its parsed state in module globals,which coupled it to shared state: the loader could not be re-run or unit tested, and its error paths left the globals dangling - a caller that continued after a failed load acted on freed or stale data, and several body descriptor strings were leaked outright.

Replace the globals with a caller-owned struct ufs_provisioning: the flash flow owns the context and its lifetime, every ufs_load() error path goes through one cleanup, and the parse helpers become static.


src/ufs.c
```diff
@@ -204,41 +224,33 @@ int ufs_load(const char *ufs_file, bool finalize_provisioning)
...
-	return 0;
+
+	if (retval)
+		ufs_provisioning_cleanup(ufs);
+
+	return retval;
 }
 
-int ufs_provisioning_execute(struct qdl_device *qdl,
+int ufs_provisioning_execute(struct ufs_provisioning *ufs, struct qdl_device *qdl,
```
quic-kdybcio, line 245: nit: this patch changes some signatures to have *ufs first, others to have *qdl first.. I think the latter is more in line with the rest of the tree

igoropaniuk: It was my deliberate choice - as it's  consistent within the ufs API: after the refactor, every `ufs_*` function is a method on the context (f.g. `ufs_provisioning_init(&ufs)`, `ufs_load(&ufs, …)`, `ufs_need_provisioning(&ufs)`, `ufs_provisioning_cleanup(&ufs)`)

tests/test_ufs.c
```diff
@@ -0,0 +1,183 @@
...
+static void test_cleanup_idempotent(void **state)
+{
+	struct ufs_provisioning ufs;
+	(void)state;
+
+	ufs_provisioning_init(&ufs);
+	/* Cleaning an untouched context is safe and repeatable. */
+	ufs_provisioning_cleanup(&ufs);
+	ufs_provisioning_cleanup(&ufs);
+	assert_empty(&ufs);
```
quic-kdybcio, line 183: it would presumably be empty, maybe a better test would be to copy ufs after the first cleanup and memcmp it after the second

igoropaniuk: makes sense, fixed

igoropaniuk:
both comments from copilot are pre-existing defects unrelated to the context of this refactor

PR #310, opened 2026-08-25
nbdkit-qdl-plugin small cleanups & documentation rework

obbardc:
When updating the Debian packaging to ship the nbdkit-qdl-plugin, I tried the built plugin with an RB3gen2 and the documentation under `doc/nbd.md`.

I found the documentation quite hard to follow, so I reworked it and while I was there I added some small quality-of-life changes to the plugin implementation along the way. The main one is the cleaner timeout when connecting to the device.

There are also some minimal future changes I'd like to add in later follow-ups, if you are interested, including:
- add a parameter to allow nbd plugin to reset firehose into a different mode on exit (e.g. edl/off/don't reset). This was mainly useful for when I was testing the plugin, e.g you don't want the device to reboot into the image after it exits, and want to reset back into EDL so you can run the plugin again. This depends on another pending change I have to add a `qdl firehose-reset <mode>` command to `qdl`; I will open that PR shortly.
- document why the plugin can't connect to device at startup
- allow matching devices based on serial number
- debug device serial number on connect and expose serial number (and other metadata) to the NBD device so userspace can read it
- unit-tests for nbdkit-qdl-plugin in CI (and before I forget; add `-Dnbdkit=enabled` into build.yml). Not sure how this can be done really; possibly could even do some full device-in-the-loop testing but I think that's probably far too much. Ideas welcome.

Let me know if you think any of these are not worth implementing.

### DIFF
--- a/src/nbdkit-qdl-plugin.c
+++ b/src/nbdkit-qdl-plugin.c
@@ -217,6 +217,9 @@ static int qdl_device_setup(void)
 		goto err_close;
 	}
 
+	nbdkit_debug("serving LUN %d; sector size %zu bytes, %zu sectors",
+		     config_lun, sector_size, num_sectors);
+
 	dev = d;
 	return 0;
 
--- a/src/nbdkit-qdl-plugin.c
+++ b/src/nbdkit-qdl-plugin.c
@@ -20,6 +20,9 @@
 
 bool qdl_debug;
 
+/* UFS supports up to eight logical units (LUNs) per device. */
+#define QDL_UFS_LUN_COUNT 8
+
 static const char *config_programmer;
 static enum qdl_storage_type config_storage = QDL_STORAGE_UFS;
 static int config_lun;
@@ -51,7 +54,12 @@ static int qdl_plugin_config(const char *key, const char *value)
 			return -1;
 		}
 	} else if (!strcmp(key, "lun")) {
-		config_lun = atoi(value);
+		if (nbdkit_parse_int("lun", value, &config_lun) == -1)
+			return -1;
+		if (config_lun < 0 || config_lun >= QDL_UFS_LUN_COUNT) {
+			nbdkit_error("lun must be between 0 and %d", QDL_UFS_LUN_COUNT - 1);
+			return -1;
+		}
 	} else if (!strcmp(key, "debug")) {
 		qdl_debug = !strcmp(value, "true") || !strcmp(value, "1");
 	} else {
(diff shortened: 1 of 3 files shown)


src/nbdkit-qdl-plugin.c
```diff
@@ -158,6 +302,18 @@ static int64_t qdl_plugin_get_size(void *handle)
 	return (int64_t)sector_size * num_sectors;
 }
 
+static int qdl_plugin_block_size(void *handle, uint32_t *minimum,
```
obbardc, line 305: This fails in CI (Ubuntu Jammy/22.04/`nbdkit 1.24.1-2ubuntu4`) with:

```
FAILED: [code=1] nbdkit-qdl-plugin.so.p/src_nbdkit-qdl-plugin.c.o 
cc -Inbdkit-qdl-plugin.so.p -I. -I.. -I../include -Isrc -I../src -I/usr/include/libusb-1.0 -I/usr/include/libxml2 -fdiagnostics-color=always -Wall -Winvalid-pch -Wextra -O2 -g -D_FILE_OFFSET_BITS=64 -fPIC -MD -MQ nbdkit-qdl-plugin.so.p/src_nbdkit-qdl-plugin.c.o -MF nbdkit-qdl-plugin.so.p/src_nbdkit-qdl-plugin.c.o.d -o nbdkit-qdl-plugin.so.p/src_nbdkit-qdl-plugin.c.o -c ../src/nbdkit-qdl-plugin.c
../src/nbdkit-qdl-plugin.c:369:10: error: ‘struct nbdkit_plugin’ has no member named ‘block_size’
  369 |         .block_size = qdl_plugin_block_size,
      |          ^~~~~~~~~~
../src/nbdkit-qdl-plugin.c:369:23: warning: initialization of ‘int (*)(void *)’ from incompatible pointer type ‘int (*)(void *, uint32_t *, uint32_t *, uint32_t *)’ {aka ‘int (*)(void *, unsigned int *, unsigned int *, unsigned int *)’} [-Wincompatible-pointer-types]
  369 |         .block_size = qdl_plugin_block_size,
      |                       ^~~~~~~~~~~~~~~~~~~~~
../src/nbdkit-qdl-plugin.c:369:23: note: (near initialization for ‘plugin.can_write’)
```

Feature was introduced in nbdkit 1.30.0 (available in ubuntu noble; debian bookworm) with https://gitlab.com/nbdkit/nbdkit/-/commit/2155ff48364a4d1a7368cf4f43ed78ce1ee47dd3

I guess we can either:
- depreciate jammy and bump ubuntu version (possibly useful to test against a newer version of libs anyway).
- gate this on nbdkit version (not sure if that's even possible?) and note in docs.
- remove this callback/feature.

Ideas welcome.

PR #314, opened 2026-08-28
firehose: give the UFS provisioning commit a longer timeout

igoropaniuk:
UFS provisioning drives the programmer with three <ufs> tags: a common tag, one or more body tags, and an epilogue. The epilogue is sent twice
- first with commit=0 to validate the requested layout, then with commit=1 to actually write the UFS configuration descriptor to the device. All of these tags went through firehose_send_single_tag(), which waited only 5s for the <response value="ACK"/>.

The commit=1 epilogue is the only expensive step: it makes the programmer write the configuration descriptor to the device, which can take considerably longer than 5s. When it does, the device emits its "Calling handler for ufs" log but has not answered with an ACK yet, so firehose_read() times out, firehose_send_single_tag() reports the tag as failed, and provisioning is aborted even though the device is still working. The program/erase write-back paths already use a 120s timeout for exactly this reason; the UFS commit needs the same treatment.

Thread a timeout into firehose_send_single_tag() and use a short 5s wait for the cheap common/body/validation tags while giving the commit epilogue a 120s window. The commit=0 validation pass keeps the short timeout.

Observed failure before the fix:

  $ qdl --storage ufs prog_firehose_ddr.elf provision_ufs31.xml --debug
  ...
  FIREHOSE WRITE: <?xml version="1.0"?>
  <data><ufs LUNtoGrow="0" commit="1"/></data>
  FIREHOSE READ: <?xml version="1.0" encoding="UTF-8" ?>
  <data>
  <log value="INFO: Calling handler for ufs" /></data>
  LOG: INFO: Calling handler for ufs
  ufs request failed
  failed to apply ufs epilogue
  UFS provisioning failed

### DIFF
--- a/src/firehose.c
+++ b/src/firehose.c
@@ -38,6 +38,18 @@ enum {
 };
 
 /*
+ * Timeouts (in milliseconds) for the UFS provisioning tags. The common and
+ * body tags only stage descriptors in the programmer, so a short wait is
+ * enough. The epilogue with commit=1 makes the programmer write the UFS
+ * configuration descriptor to the device, which can take considerably
+ * longer than the previous 5s window - matching the write-back timeout used
+ * by the program/erase paths avoids spuriously reporting a failure while the
+ * device is still provisioning.
+ */
+#define FIREHOSE_UFS_TAG_TIMEOUT_MS	5000
+#define FIREHOSE_UFS_COMMIT_TIMEOUT_MS	120000
+
+/*
  * Substring emitted by the Firehose programmer's startup logs when VIP is
  * active on-device (e.g. "INFO: VIP is enabled, receiving the signed table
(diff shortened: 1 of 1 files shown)


quic-kdybcio (approved):

JohnSagaQuic:
I reviewed firehose.c and noticed a possible issue in firehose_apply_ufs_epilogue(), line 1517.
The function waits up to 120 seconds for an ACK. If the device resets the USB connection before the ACK is received, firehose_read() returns -EIO, which causes the read loop to exit. As a result, firehose_send_single_tag() returns a non-ACK value, leading firehose_apply_ufs_epilogue() to report "failed to apply ufs epilogue" and return -1, even though UFS provisioning may have completed successfully.
Would it make sense to treat -EIO during the commit epilogue response as a potential success case, since a device reset at that stage may be expected?

PR #321, opened 2026-09-11
ux: add consistent timestamps to debug diagnostics

igoropaniuk:
Add millisecond timestamps to diagnostic messages when `--debug` is enabled, making it easier to identify stalls, investigate timeouts, and correlate host and device activity.

Route remaining warnings, fatal errors, and hex dumps through shared logging for consistent timestamp coverage. Preserve non-debug output, errno details, and exit statuses.


tests/test_programmer.c
```diff
@@ -42,14 +42,12 @@ int decode_sahara_config(struct sahara_image *blob, struct sahara_image *images,
 	return 0;
 }
 
-#ifdef _WIN32
-/* err() lives in oscompat.c, which the test does not link. */
-void err(int eval, const char *fmt, ...)
+/* Fatal logging lives in ux.c, which the test does not link. */
+void ux_die(int eval, const char *fmt, ...)
```
quic-kdybcio, line 46: let's keep the noreturn attribute here too,

